PR #1, opened 2024-01-31
Init with repo-config

Marenz:
- Regenerate repository using repo-config
- Remove cookiecutter todos


README.md
```diff
@@ -8,8 +8,6 @@
 
 Dispatch API client for Python
 
-TODO(cookiecutter): Improve the README file
-
```
llucax: Not sure if removing this one is really fair, the README is not really improved :P

mkdocs.yml
```diff
@@ -8,15 +8,12 @@ site_author: "Frequenz Energy-as-a-Service GmbH"
 copyright: "Copyright © 2024 Frequenz Energy-as-a-Service GmbH"
 repo_name: "frequenz-client-dispatch-python"
 repo_url: "https://github.com/frequenz-floss/frequenz-client-dispatch-python"
-# TODO(cookiecutter): "main" is the GitHub repo default branch, you might want to update it
-# if the project uses a different default branch.
 edit_uri: "edit/main/docs/"
```
llucax: Strange, I thought I updated this to use v0.x.x instead. You should probably update it (and I should update the default branch to be named like this, I will do it now).

mkdocs.yml
```diff
@@ -116,8 +112,6 @@ plugins:
             show_source: true
             signature_crossrefs: true
           import:
-            # TODO(cookiecutter): You might want to add other external references here
-            # See https://mkdocstrings.github.io/python/usage/#import for details
```
llucax: You should add `frequenz-api-dispatch` here, so we get proper cross-reference.

pyproject.toml
```diff
@@ -26,7 +25,6 @@ classifiers = [
   "Typing :: Typed",
 ]
 requires-python = ">= 3.11, < 4"
-# TODO(cookiecutter): Remove and add more dependencies if appropriate
```
llucax: Oh, OK, the dependency to `frequenz-api-dispatch` is not added yet. Ignore the comment about the cross-reference if you don't plan to add it in this PR.

llucax (commented):
I think some TODOs shouldn't be removed just yet, as they are not really addressed.

pyproject.toml
Marenz: I suppose I should do it in this PR else i might forget all these places to reference it

llucax: Yeah, I agree it makes sense to include the actual work to remove the TODOs in this initial PR. :+1:

README.md
Marenz: I'd argue that it's better than when it contained a note that it needs to be improved :P

mkdocs.yml
Marenz: Changed, but I can't update it for this repo (no permission)

llucax: Renamed, please remember to update your local branch name.

README.md
llucax: <p>I don't agree, it is worse because now it won't remind you that it needs improvement 
think
</p>

mkdocs.yml
```diff
@@ -0,0 +1,131 @@
...
+            show_category_heading: true
+            show_root_heading: true
+            show_root_members_full_path: true
+            show_signature_annotations: true
+            show_source: true
+            signature_crossrefs: true
+          import:
+            - https://docs.python.org/3/objects.inv
+            - https://typing-extensions.readthedocs.io/en/stable/objects.inv
+            - https://frequenz-floss.github.io/frequenz-api-dispatch
```
llucax: This should be:
```suggestion
            - https://frequenz-floss.github.io/frequenz-api-dispatch/v0.12/objects.inv
```

llucax (commented):
The commit message "Handle or remove cookiecutter todos" sounds very weird to me, I would just merge it with the commit removing the TODOs or split it in one commit per "fixed" todo, and I would keep the readme TODO, but I consider these minor, so Iḿ fine to approve with those as they are.

mkdocs.yml
Marenz: set to a fixed version? :/

README.md
Marenz: Every time you see it it's pretty clear it needs improvement :P

Marenz:
> I would just merge it with the commit

That was the actual plan.

Updated mkdocs to have the right reference

mkdocs.yml
llucax: Yes, for doing cross-referencing we need to specify a version, as it needs to obtain a list of symbols to do the linking. It's not ideal, specially when we start to have proper minors, as links will always send you to the minimum supported minor version, but I don't see a way around this.

llucax (approved):

mkdocs.yml
Marenz: This place is also easily forgotten to be updated..

PR #14, opened 2024-02-05
Create initial structure

Marenz:
pylint still makes some problems, apparently it can't find the generated files?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -230,7 +230,7 @@ jobs:
         run: python -m build
 
       - name: Upload distribution files
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: dist-packages
           path: dist/
@@ -294,7 +294,7 @@ jobs:
           mike set-default $MIKE_VERSION
 
       - name: Upload site
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: docs-site
           path: site/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,9 +26,17 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "typing-extensions >= 4.5.0, < 5",
-  "frequenz-api-dispatch >= 0.12.0, < 0.13",
-  "frequenz-client-base >= 0.1.0, < 0.2",
+  "typing-extensions >= 4.6.1, < 5",
+  # Use direct commit until release
+  "frequenz-api-dispatch @ git+https://github.com/frequenz-floss/frequenz-api-dispatch.git@223315c",
+#  "frequenz-api-dispatch >= 0.13.0, < 0.14",
+  # Directly use unreleased commit until the next release
+  "frequenz-client-base @ git+https://github.com/frequenz-floss/frequenz-client-base-python.git@57d3093",
+#  "frequenz-client-base >= 0.1.0, < 0.2",
+  # Directly use unreleased commit until the first release
+  "frequenz-client-common @ git+https://github.com/frequenz-floss/frequenz-client-common-python.git@5e6e7b7",
+#  "frequenz-client-common == 0.1.0",
+  "grpcio >= 1.54.2, < 2"
 ]
 dynamic = ["version"]
 
@@ -61,6 +69,7 @@ dev-mypy = [
   "types-Markdown == 3.4.2.10",
   # For checking the noxfile, docs/ script, and tests
   "frequenz-dispatch-client[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "grpc-stubs == 1.53.0.2",
 ]
 dev-noxfile = [
   "nox == 2023.4.22",
@@ -70,6 +79,7 @@ dev-pylint = [
   "pylint == 3.0.2",
   # For checking the noxfile, docs/ script, and tests
   "frequenz-dispatch-client[dev-mkdocs,dev-noxfile,dev-pytest]",
+  "frequenz-api-dispatch @ git+https://github.com/frequenz-floss/frequenz-api-dispatch.git@223315c",
 ]
 dev-pytest = [
   "pytest == 7.4.2",
@@ -77,6 +87,7 @@ dev-pytest = [
   "pytest-mock == 3.11.1",
   "pytest-asyncio == 0.21.1",
   "async-solipsism == 0.5",
+  "hypothesis == 6.98.8",
 ]
 dev = [
   "frequenz-dispatch-client[dev-mkdocs,dev-flake8,dev-formatting,dev-mkdocs,dev-mypy,dev-noxfile,dev-pylint,dev-pytest]",
--- a/src/frequenz/client/dispatch/_client.py
+++ b/src/frequenz/client/dispatch/_client.py
@@ -1,0 +1,257 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Dispatch API client for Python."""
+from datetime import datetime, timedelta
+from typing import Any, AsyncIterator, Iterator
+
+import grpc
+from frequenz.api.dispatch.v1 import dispatch_pb2_grpc
+
+# pylint: disable=no-name-in-module
+from frequenz.api.dispatch.v1.dispatch_pb2 import (
+    DispatchDeleteRequest,
+    DispatchFilter,
+    DispatchGetRequest,
+    DispatchListRequest,
+    DispatchUpdateRequest,
+)
+from frequenz.api.dispatch.v1.dispatch_pb2 import (
+    TimeIntervalFilter as PBTimeIntervalFilter,
+)
+from google.protobuf.timestamp_pb2 import Timestamp
+
+from ._internal_types import DispatchCreateRequest
+from .types import (
+    ComponentSelector,
+    Dispatch,
+    RecurrenceRule,
+    component_selector_to_protobuf,
+)
+
+# pylint: enable=no-name-in-module
+
+
+class Client:
+    """Dispatch API client."""
+
+    def __init__(self, grpc_channel: grpc.aio.Channel, svc_addr: str) -> None:
+        """Initialize the client.
+
+        Args:
+            grpc_channel: gRPC channel to use for communication with the API.
+            svc_addr: Address of the service to connect to.
+        """
+        self._svc_addr = svc_addr
+        self._stub = dispatch_pb2_grpc.MicrogridDispatchServiceStub(grpc_channel)
+
+    # pylint: disable=too-many-arguments, too-many-locals
+    async def list(
+        self,
+        microgrid_id: int,
+        component_selectors: Iterator[ComponentSelector] = iter(()),
+        start_from: datetime | None = None,
+        start_to: datetime | None = None,
+        end_from: datetime | None = None,
+        end_to: datetime | None = None,
+        active: bool | None = None,
+        dry_run: bool | None = None,
+    ) -> AsyncIterator[Dispatch]:
+        """List dispatches.
+
+        Example usage:
+
+        ```python
+        grpc_channel = grpc.aio.insecure_channel("example")
+        client = Client(grpc_channel, "localhost:50051")
+        async for dispatch in client.list(microgrid_id=1):
+            print(dispatch)
+        ```
+
+        Yields:
+            Dispatch: The dispatches.
+
+        Args:
+            microgrid_id: The microgrid_id to list dispatches for.
+            component_selectors: optional, list of component ids or categories to filter by.
+            start_from: optional, filter by start_time >= start_from.
+            start_to: optional, filter by start_time < start_to.
+            end_from: optional, filter by end_time >= end_from.
+            end_to: optional, filter by end_time < end_to.
+            active: optional, filter by active status.
+            dry_run: optional, filter by dry_run status.
+
+        Returns:
+            An async iterator of dispatches.
+        """
+        time_interval = None
+
+        def to_timestamp(dt: datetime | None) -> Timestamp | None:
+            if dt is None:
+                return None
+
+            ts = Timestamp()
+            ts.FromDatetime(dt)
+            return ts
+
+        if start_from or start_to or end_from or end_to:
+            time_interval = PBTimeIntervalFilter(
+                start_from=to_timestamp(start_from),
+                start_to=to_timestamp(start_to),
+                end_from=to_timestamp(end_from),
+                end_to=to_timestamp(end_to),
+            )
+
+        selectors = []
+
+        for selector in component_selectors:
+            selectors.append(component_selector_to_protobuf(selector))
+
+        filters = DispatchFilter(
+            selectors=selectors,
+            time_interval=time_interval,
+            is_active=active,
+            is_dry_run=dry_run,
+        )
+        request = DispatchListRequest(microgrid_id=microgrid_id, filter=filters)
+
+        response = await self._stub.ListMicrogridDispatches(request)  # type: ignore
+        for dispatch in response.dispatches:
+            yield Dispatch.from_protobuf(dispatch)
+
+    async def create(
+        self,
+        microgrid_id: int,
+        _type: str,
+        start_time: datetime,
+        duration: timedelta,
+        selector: ComponentSelector,
+        active: bool = True,
+        dry_run: bool = False,
+        payload: dict[str, Any] | None = None,
+        recurrence: RecurrenceRule | None = None,
+    ) -> None:
+        """Create a dispatch.
+
+        Args:
+            microgrid_id: The microgrid_id to create the dispatch for.
+            _type: User defined string to identify the dispatch type.
+            start_time: The start time of the dispatch.
+            duration: The duration of the dispatch.
+            selector: The component selector for the dispatch.
+            active: The active status of the dispatch.
+            dry_run: The dry_run status of the dispatch.
+            payload: The payload of the dispatch.
+            recurrence: The recurrence rule of the dispatch.
+
+        Raises:
+            ValueError: If start_time is in the past.
+        """
+        if start_time <= datetime.now().astimezone(start_time.tzinfo):
+            raise ValueError("start_time must not be in the past")
+
+        request = DispatchCreateRequest(
+            microgrid_id=microgrid_id,
+            type=_type,
+            start_time=start_time,
+            duration=duration,
+            selector=selector,
+            is_active=active,
+            is_dry_run=dry_run,
+            payload=payload or {},
+            recurrence=recurrence or RecurrenceRule(),
+        ).to_protobuf()
+
+        await self._stub.CreateMicrogridDispatch(request)  # type: ignore
+
+    async def update(
+        self,
+        dispatch_id: int,
+        new_fields: dict[str, Any],
+    ) -> None:
+        """Update a dispatch.
+
+        The `new_fields` argument is a dictionary of fields to update. The keys are
+        the field names, and the values are the new values for the fields.
+
+        For recurrence fields, the keys are preceeded by "recurrence.".
+
+        Args:
+            dispatch_id: The dispatch_id to update.
+            new_fields: The fields to update.
+        """
+        msg = DispatchUpdateRequest(id=dispatch_id)
+
+        for key, val in new_fields.items():
+            path = key.split(".")
+
+            match path[0]:
+                case "type":
+                    msg.update.type = val
+                case "start_time":
+                    msg.update.start_time.FromDatetime(val)
+                case "duration":
+                    msg.update.duration = int(val.total_seconds())
+                case "selector":
+                    msg.update.selector.CopyFrom(component_selector_to_protobuf(val))
+                case "is_active":
+                    msg.update.is_active = val
+                case "active":
+                    msg.update.is_active = val
+                    key = "is_active"
+                case "is_dry_run":
+                    msg.update.is_dry_run = val
+                case "dry_run":
+                    msg.update.is_dry_run = val
+                    key = "is_dry_run"
+                case "recurrence":
+                    match path[1]:
+                        case "freq":
+                            msg.update.recurrence.freq = val
+                        # Proto uses "freq" instead of "frequency"
+                        case "frequency":
+                            msg.update.recurrence.freq = val
+                            # Correct the key to "recurrence.freq"
+                            key = "recurrence.freq"
+                        case "interval":
+                            msg.update.recurrence.interval = val
+                        case "end_criteria":
+                            msg.update.recurrence.end_criteria.CopyFrom(
+                                val.to_protobuf()
+                            )
+                        case "byminutes":
+                            msg.update.recurrence.byminutes.extend(val)
+                        case "byhours":
+                            msg.update.recurrence.byhours.extend(val)
+                        case "byweekdays":
+                            msg.update.recurrence.byweekdays.extend(val)
+                        case "bymonthdays":
+                            msg.update.recurrence.bymonthdays.extend(val)
+                        case "bymonths":
+                            msg.update.recurrence.bymonths.extend(val)
+
+            msg.update_mask.paths.append(key)
+
+        await self._stub.UpdateMicrogridDispatch(msg)  # type: ignore
+
+    async def get(self, dispatch_id: int) -> Dispatch:
+        """Get a dispatch.
+
+        Args:
+            dispatch_id: The dispatch_id to get.
+
+        Returns:
+            Dispatch: The dispatch.
+        """
+        request = DispatchGetRequest(id=dispatch_id)
+        response = await self._stub.GetMicrogridDispatch(request)  # type: ignore
+        return Dispatch.from_protobuf(response)
+
+    async def delete(self, dispatch_id: int) -> None:
+        """Delete a dispatch.
+
+        Args:
+            dispatch_id: The dispatch_id to delete.
+        """
+        request = DispatchDeleteRequest(id=dispatch_id)
+        await self._stub.DeleteMicrogridDispatch(request)  # type: ignore
--- a/tests/test_dispatch_client.py
+++ b/tests/test_dispatch_client.py
@@ -2,4 +2,472 @@
 # Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
 """Tests for the frequenz.client.dispatch package."""
-import pytest
+
+import random
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from typing import Any
+
+import grpc
+
+# pylint: disable=no-name-in-module
+from frequenz.api.dispatch.v1.dispatch_pb2 import Dispatch as PBDispatch
+from frequenz.api.dispatch.v1.dispatch_pb2 import (
+    DispatchCreateRequest as PBDispatchCreateRequest,
+)
+from frequenz.api.dispatch.v1.dispatch_pb2 import (
+    DispatchDeleteRequest,
+    DispatchGetRequest,
+    DispatchList,
+)
+from frequenz.api.dispatch.v1.dispatch_pb2 import (
+    DispatchListRequest as PBDispatchListRequest,
+)
+from frequenz.api.dispatch.v1.dispatch_pb2 import DispatchUpdateRequest
+from frequenz.client.common.microgrid.components import ComponentCategory
+from google.protobuf.empty_pb2 import Empty
+from google.protobuf.timestamp_pb2 import Timestamp
+from pytest import raises
+
+from frequenz.client.dispatch import Client
+from frequenz.client.dispatch._internal_types import DispatchCreateRequest
+from frequenz.client.dispatch.types import (
+    Dispatch,
+    EndCriteria,
+    Frequency,
+    RecurrenceRule,
+    Weekday,
+)
+
+
+class DispatchSampler:
+    """Generates random dispatch messages."""
+
+    def __init__(self, seed: int = 0) -> None:
+        """Initialize the sampler.
+
+        Args:
+            seed: seed to initialize the rng with
+        """
+        self._rng = random.Random(seed)
+        self._last_id: int = 0
+
+    def recurrence_rule(self) -> RecurrenceRule:
+        """Generate a random recurrence rule.
+
+        Returns:
+            a random recurrence rule
+        """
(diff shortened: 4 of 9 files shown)


Marenz:
@llucax is there a trick to make pylint happy or do I need to add exceptions for most of those?

```
nox > pylint src docs noxfile.py tests
************* Module frequenz.client.dispatch._client
# This one is clear
src/frequenz/client/dispatch/_client.py:100:9: W0511: TODO handle result (fixme)
src/frequenz/client/dispatch/_client.py:11:0: E0611: No name 'Timestamp' in module 'google.protobuf.timestamp_pb2' (no-name-in-module)
src/frequenz/client/dispatch/_client.py:72:28: E1101: Module 'frequenz.api.dispatch.v1.dispatch_pb2' has no 'TimeIntervalFilter' member (no-member)
src/frequenz/client/dispatch/_client.py:83:33: E1101: Module 'frequenz.api.dispatch.v1.dispatch_pb2' has no 'ComponentSelector' member (no-member)
src/frequenz/client/dispatch/_client.py:90:18: E1101: Module 'frequenz.api.dispatch.v1.dispatch_pb2' has no 'DispatchFilter' member (no-member)
src/frequenz/client/dispatch/_client.py:96:18: E1101: Module 'frequenz.api.dispatch.v1.dispatch_pb2' has no 'DispatchListRequest' member (no-member)
************* Module test_dispatch_client
# This one is clear
tests/test_dispatch_client.py:9:4: C0415: Import outside toplevel (frequenz.client.dispatch.Client) (import-outside-toplevel)
```

llucax:
It is a known bug, we need to put an ignore in the import line:

* https://github.com/pylint-dev/pylint/issues/6281

pyproject.toml
```diff
@@ -26,9 +26,14 @@ classifiers = [
...
-  "frequenz-api-dispatch >= 0.12.0, < 0.13",
-  "frequenz-client-base >= 0.1.0, < 0.2",
+  "typing-extensions >= 4.6.1, < 5",
+# Use marenz fork @ 0d04bc3 until the next release
+  "frequenz-api-dispatch @ git+https://github.com/marenz/frequenz-api-dispatch.git@0d04bc3",
+#  "frequenz-api-dispatch >= 0.12.0, < 0.13",
+# Directly use unreleased/unmerged commit fe4e238 until the next release
+  "frequenz-client-base @ git+https://github.com/marenz/frequenz-client-base-python.git@fe4e238",
+#  "frequenz-client-base >= 0.1.0, < 0.2",
+  "frequenz-sdk == v1.0.0-rc4",
```
llucax: We shouldn't import the SDK in API clients, the idea is to do it the other way around (the SDK will import the client).

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,101 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Dispatch API client for Python."""
+from dataclasses import dataclass
+from datetime import datetime
+
+import grpc
+from frequenz.api.dispatch.v1 import dispatch_pb2, dispatch_pb2_grpc
+from frequenz.sdk.microgrid.component import ComponentCategory
```
llucax: This should come from https://github.com/frequenz-floss/frequenz-client-common-python/. We still need to move/recreate the wrappers in that repo (they are not available yet).

@frequenz-floss/api-reporting-team @frequenz-floss/api-microgrid-team FYI as this is common to at least all these APIs

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,101 @@
...
+from google.protobuf.timestamp_pb2 import Timestamp
+
+# from ._types import ForecastFeature, Forecasts, Location
+
+
+@dataclass
```
llucax:
```suggestion
@dataclass(frozen=True, kw_only=True)
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,101 @@
...
+            ts.FromDatetime(dt)
+            return ts
+
+        if interval_filter:
+            time_interval = dispatch_pb2.TimeIntervalFilter(
+                start_from=to_timestamp(interval_filter.start_from),
+                start_to=to_timestamp(interval_filter.start_to),
+                end_from=to_timestamp(interval_filter.end_from),
+                end_to=to_timestamp(interval_filter.end_to),
+            )
```
llucax: The idea is that your `TimeIntervalFilter` wrapper class has `from_proto()` and `to_proto()` methods. For now in the SDK we only have `from_proto()` ([example](https://github.com/frequenz-floss/frequenz-sdk-python/blob/91821a3e2719024c0b2a114d6a11aa692d3429d6/src/frequenz/sdk/microgrid/component/_component_data.py#L85-L112)) but I guess it could make sense to have the other too.

That said, to be honest, I don't like this approach, as the idea is to hide the underlying protocol we are using in the client, the idea is that the client user doesn't need to know about protobuf. Since conversion only needs to happen in the client internally, I would prefer an approach where we provide conversion functions separately from the wrapper class, similar to what marshmallow does, or even dataclass when converting to `dict` or `tuple`.

Maybe this is something to discuss in the next API meeting (or even SDK meeting, as this is a Python thing more than a API thing).

FYI @frequenz-floss/python-sdk-team

tests/test_dispatch_client.py
```diff
@@ -2,4 +2,10 @@
 # Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
 """Tests for the frequenz.client.dispatch package."""
-import pytest
+
+
+def test_client() -> None:
+    """Test the dispatch client."""
```
llucax: That's a bold statement! :D

llucax (commented):
Hi, some early feedback. About the SDK thing, if it is only a temporary hack to get things going, I'm good with it!

src/frequenz/client/dispatch/_client.py
Marenz: I am not sure for this particular type it's worth having all that, given that it's only used in this function and no-where outside, and the conversion is extremely simple

llucax: Right, maybe I picked the wrong example.

tests/test_dispatch_client.py
Marenz: give it time :P

.github/workflows/ci.yaml
llucax, line 1: This should be split to a different commit.

src/frequenz/client/dispatch/_types.py
```diff
@@ -0,0 +1,348 @@
...
+)
+
+ComponentSelector = list[int] | ComponentCategory
+"""A component selector specifying which components a dispatch targets.
+
+A component selector can be a list of component IDs or a component category.
+"""
+
+
+def _component_selector_from_protobuf(
```
llucax: I would make this and the `to_protobuf` functions public (not starting with `_`) as you are importing it in another module. This whole module is private so it's not like the end user will have (direct) access to it.

src/frequenz/client/dispatch/_types.py
```diff
@@ -0,0 +1,348 @@
...
+        pb_dispatch.microgrid_id = self.microgrid_id
+        pb_dispatch.type = self.type
+        pb_dispatch.create_time.FromDatetime(self.create_time)
+        pb_dispatch.update_time.FromDatetime(self.update_time)
+        pb_dispatch.start_time.FromDatetime(self.start_time)
+        pb_dispatch.duration = int(self.duration.total_seconds())
+        pb_dispatch.selector.CopyFrom(_component_selector_to_protobuf(self.selector))
+        pb_dispatch.is_active = self.is_active
+        pb_dispatch.is_dry_run = self.is_dry_run
+        # pb_dispatch.payload = self.payload
```
llucax: Why is this commented out?

tests/test_dispatch_client.py
```diff
@@ -2,4 +2,97 @@
...
+    dispatch_servicer = DispatchServicer()
+    dispatch_pb2_grpc.add_MicrogridDispatchServiceServicer_to_server(
+        dispatch_servicer, server
+    )
+    port = server.add_insecure_port("localhost:0")
+    await server.start()
+    try:
+        yield server, f"localhost:{port}"
+    finally:
+        await server.stop(None)
```
llucax: I should **not** use this mock service if it is really listening to a port, we did that in the SDK and it had a lot of issues, among them that multiple tests can be run at the same time. We should mock the server instead at the python level, patching the functions we are calling in our code.

All you need is to mock/patch (in the pyhon sense) this function used by the client:  `self._stub.ListMicrogridDispatches(request)` to return whatever mock data you want to have. We don't need to test the whole gRPC stuff, only our client code.

We could also do a test of the whole thing, but that should be an integration test, not a unit test.

tests/test_dispatch_client.py
```diff
@@ -2,4 +2,97 @@
...
+
+
+@pytest.mark.asyncio
```
llucax: This mark shouldn't be necessary, usually pytest detects it automatically. Any reason to include it?

tests/test_dispatch_types.py
```diff
@@ -0,0 +1,160 @@
...
+
+    for selector in (
+        [1, 2, 3],
+        [10, 20, 30],
+        ComponentCategory.BATTERY,
+        ComponentCategory.GRID,
+        ComponentCategory.METER,
+    ):
+        protobuf = _component_selector_to_protobuf(selector)  # type: ignore
+        assert _component_selector_from_protobuf(protobuf) == selector
```
llucax: Any reason not to use `pytest.mark.parametrize` for this? It will move some clutter out of the test itself.

Maybe you can borrow some ideas from here if you have more complex tests and you want, for example, to assign names to them, etc.:

https://github.com/frequenz-floss/pymdownx-superfence-filter-lines-python/blob/231d34c4f80ad00785b231d49e28453b5ccfb08e/tests/test_do_validate.py#L33-L127

tests/test_dispatch_types.py
```diff
@@ -0,0 +1,160 @@
...
+def test_component_selector() -> None:
+    """Test the component selector."""
+    # pylint: disable=import-outside-toplevel
+    from frequenz.client.common.microgrid.components.components import ComponentCategory
+
+    from frequenz.client.dispatch._types import (
+        ComponentSelector,
+        _component_selector_from_protobuf,
+        _component_selector_to_protobuf,
+    )
```
llucax: Any reason for the imports inside the test?

tests/test_dispatch_types.py
```diff
@@ -0,0 +1,160 @@
...
+
+    assert ComponentSelector is not None
```
llucax: Why this assert? How could it be `None`? I don't get this.

I guess maybe is a copy&paste thing from the very silly tests in the generated files in the API? In there it makes *some* sense because we are "testing" generated code, and generated code that is super dynamic, so maybe it could happen that you import something and it is `None`, but not even sure if they make sense there either.

llucax (commented):
Mostly minor things, except for the gRPC mock server which caused us a lot of headaches in the SDK.

src/frequenz/client/dispatch/_types.py
Marenz: It caused trouble because the attribute is read-only. Haven'- investigated more

tests/test_dispatch_client.py
Marenz: It was ChatGPTed :angel:

tests/test_dispatch_types.py
Marenz: Tbh, I didn't feel the need to do more complex testing with these methods, they seem simple enough

tests/test_dispatch_types.py
Marenz: I guess keeping things near where they are used. No deeper reason

src/frequenz/client/dispatch/_types.py
llucax: Oh, weird. Maybe add a comment or create an issue about it?

tests/test_dispatch_types.py
llucax: If there is no good reason, I would put it on the top-level, which is the most popular style and doesn't require an exception for `pylint`.

tests/test_dispatch_types.py
llucax: Sure, you don't have to make it more complicated, but using `parametrize()` is actually simpler, you don't need to create classes for test cases, that was just a FYI in case it was useful. Just using parametrize is as simple as:

```py
from frequenz.client.common.microgrid.components.components import ComponentCategory

from frequenz.client.dispatch._types import (
    ComponentSelector,
    _component_selector_from_protobuf,
    _component_selector_to_protobuf,
)

@pytest.mark.parametrize("selector", [
        [1, 2, 3],
        [10, 20, 30],
        ComponentCategory.BATTERY,
        ComponentCategory.GRID,
        ComponentCategory.METER,
    ])
def test_component_selector(selector: list[int] | ComponentCategory) -> None:
    """Test the component selector."""
    protobuf = _component_selector_to_protobuf(selector)  # type: ignore
    assert _component_selector_from_protobuf(protobuf) == selector
```

Another advantage of using parametrize is it creates one test per parameter, so you get exactly which parameter failed and you can execute the test for one particular parameter only, also without it, the test will be aborted after the first failure, while using parametrize will continue testing other parameters (unless the fail early option is used).

If you think updating these tests here now are not worth it, I'm OKish with that, but I strongly suggest using `parametrize()` in future tests.

Marenz:
@llucax this is almost ready, some mypy/pylint/flake8 complains, but everything else is working already. Feel free to review already.

pyproject.toml
```diff
@@ -26,9 +26,17 @@ classifiers = [
...
+  "typing-extensions >= 4.6.1, < 5",
+# Use direct commit until release
+  "frequenz-api-dispatch @ git+https://github.com/frequenz-floss/frequenz-api-dispatch.git@223315c",
+#  "frequenz-api-dispatch >= 0.13.0, < 0.14",
+# Directly use unreleased commit until the next release
+  "frequenz-client-base @ git+https://github.com/frequenz-floss/frequenz-client-base-python.git@57d3093",
+#  "frequenz-client-base >= 0.1.0, < 0.2",
+# Directly use unreleased/unmerged commit until the first release
+  "frequenz-client-common @ git+https://github.com/frequenz-floss/frequenz-client-common-python.git@5e6e7b7",
+#  "frequenz-client-common == 0.1.0",
```
llucax: This is 100% the OCD in me talking, but any reason why the comments are not indented the same as the other lines? :grimacing:

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # pylint: disable=too-many-arguments, too-many-locals
+    async def list(
+        self,
+        microgrid_id: int,
+        component_selectors: list[ComponentSelector] | None = None,
+        start_from: datetime | None = None,
+        start_to: datetime | None = None,
+        end_from: datetime | None = None,
+        end_to: datetime | None = None,
```
llucax, line 56: Not necessarily for this PR, but maybe in the future we want to add a `DatetimeRange` dataclass to pass pairs `from`/`to`.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        is_active: bool | None = None,
+        is_dry_run: bool | None = None,
```
llucax: Minor: I think attributes are more readable as `is_xxx`, but arguments read better using just `xxx` (so just `active` and `dry_run` in this case). I've taken [this recommendation from dart's style guide](https://dart.dev/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable) and I find them very good and the resulting code very readable.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        time_interval = None
+
+        def to_timestamp(dt: datetime | None) -> Timestamp | None:
+            if dt is None:
+                return None
+
+            ts = Timestamp()
+            ts.FromDatetime(dt)
+            return ts
```
llucax, line 95: This might be useful to include in https://github.com/frequenz-floss/frequenz-client-base-python/, as all repos dealing with `datetime` will have to do it.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+
+        if start_from or start_to or end_from or end_to:
+            time_interval = PBTimeIntervalFilter(
+                start_from=to_timestamp(start_from),
+                start_to=to_timestamp(start_to),
+                end_from=to_timestamp(end_from),
+                end_to=to_timestamp(end_to),
+            )
```
llucax, line 103: Just out of curiosity, this is some sort of optimization to send a message as small as possible over the wire, right? But it should be also possible to always create the PB object with all `None`s and it will have the same effect?

If so maybe you can add a comment to clarify it, so nobody thinks it is redundant in the future and removes the `if`.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+
+        selectors = []
+
+        for selector in component_selectors if component_selectors else []:
```
llucax: You can shorten this a bit with:

```suggestion
        for selector in component_selectors or []:
```

But it is probably even simpler to make the argument a tuple and have a `()` default instead.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+            svc_addr: Address of the service to connect to.
+        """
+        self._svc_addr = svc_addr
+        self._stub = dispatch_pb2_grpc.MicrogridDispatchServiceStub(grpc_channel)
+
+    # pylint: disable=too-many-arguments, too-many-locals
+    async def list(
+        self,
+        microgrid_id: int,
+        component_selectors: list[ComponentSelector] | None = None,
```
llucax:
```suggestion
        component_selectors: Iterator[ComponentSelector] = (),
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        start_from: datetime | None = None,
+        start_to: datetime | None = None,
+        end_from: datetime | None = None,
+        end_to: datetime | None = None,
+        is_active: bool | None = None,
+        is_dry_run: bool | None = None,
+    ) -> list[Dispatch]:
```
llucax: I think this should return an async iterator instead.
```suggestion
    ) -> AsyncIterator[Dispatch]:
```

So users can just do:

```py
async for dispatch in client.list():
    ...
```

Once the return type is `AsyncIterator`, you can just `yield` each value in the function body. This will be particularly important when we add pagination, so users don't need to wait (and hold in memory) all the pages before starting to process dispatches.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+            selectors=selectors,
+            time_interval=time_interval,
+            is_active=is_active,
+            is_dry_run=is_dry_run,
+        )
+        request = DispatchListRequest(microgrid_id=microgrid_id, filter=filters)
+
+        response: Awaitable[DispatchList]
+        response = self._stub.ListMicrogridDispatches(request)  # type: ignore
+        return list(map(Dispatch.from_protobuf, (await response).dispatches))
```
llucax:
```suggestion
        response = await self._stub.ListMicrogridDispatches(request)
        for dispatch in response.dispatches:
            yield Dispatch.from_protobuf(dispatch)
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        self,
+        microgrid_id: int,
+        _type: str,
+        start_time: datetime,
+        duration: timedelta,
+        selector: ComponentSelector,
+        is_active: bool,
+        is_dry_run: bool,
+        payload: dict[str, Any],
+        recurrence: RecurrenceRule,
```
llucax:
```suggestion
        recurrence: RecurrenceRule | None = None,
```
?

In this case I'm pretty sure recurrence rules are optional, but we should also document it in the protobuf spec.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        response = self._stub.ListMicrogridDispatches(request)  # type: ignore
+        return list(map(Dispatch.from_protobuf, (await response).dispatches))
+
+    async def create(
+        self,
+        microgrid_id: int,
+        _type: str,
+        start_time: datetime,
+        duration: timedelta,
+        selector: ComponentSelector,
```
llucax, line 128:
```suggestion
        selector: ComponentSelector | None = None,
```
? (and move it to the end)

Every sub-message in protobuf is implicitly `optional`. Unless the server will check its existence and fail if it is not provided. I checked the docs and it doesn't seem to say what happens if there is no selector, but I would assume all dispatches will be returned.

@stefan-brus-frequenz can you confirm this? We should also add it to the proto documentation.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+            selector: The component selector for the dispatch.
+            is_active: The is_active status of the dispatch.
+            is_dry_run: The is_dry_run status of the dispatch.
+            payload: The payload of the dispatch.
+            recurrence: The recurrence rule of the dispatch.
+        """
+        request = DispatchCreateRequest(
+            microgrid_id=microgrid_id,
+            type=_type,
+            start_time=start_time,
```
llucax, line 156: From https://github.com/frequenz-floss/frequenz-api-dispatch/blob/223315c6038778acf0b2384f881c69155d96a49c/proto/frequenz/api/dispatch/v1/dispatch.proto#L324C1-L328C44, we should check for this:

```proto
  // The start time
  // When creating a dispatch, ensure that the starting timestamp is set to
  // the current time or any future time.
  // Timestamps earlier than the current time are not allowed.
  google.protobuf.Timestamp start_time = 3;
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+    async def create(
+        self,
+        microgrid_id: int,
+        _type: str,
+        start_time: datetime,
+        duration: timedelta,
+        selector: ComponentSelector,
+        is_active: bool,
+        is_dry_run: bool,
+        payload: dict[str, Any],
```
llucax:
```suggestion
        payload: dict[str, Any] | None = None,
```
? Also pretty sure the payload is optional.

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+
+    async def create(
+        self,
+        microgrid_id: int,
+        _type: str,
+        start_time: datetime,
+        duration: timedelta,
+        selector: ComponentSelector,
+        is_active: bool,
+        is_dry_run: bool,
```
llucax: I'd probably also make these optional, defaulting to `True` and `False` respectively.

@stefan-brus-frequenz any opinion on this? Since these are not `optional` in protobuf, if omitted they will default to something, which I would guess is `False`?

src/frequenz/client/dispatch/_client.py
```diff
@@ -0,0 +1,234 @@
...
+            payload=payload,
+            recurrence=recurrence,
+        ).to_protobuf()
+
+        await self._stub.CreateMicrogridDispatch(request)  # type: ignore
+
+    async def update(
+        self,
+        dispatch_id: int,
+        new_fields: dict[str, Any],
```
llucax, line 170:
```suggestion
        **new_fields: Any,
```

But I would actually list explicitly everything that can be updated here, otherwise we lose type safety and also we allow introducing bugs because of typos (`update(start_tim=...)` won't be caught by `mypy`) and the code is quite convoluted with the huge `match` statement. Actually now that I see it the `**new_fields: Any` won't work because you use `.` in keys.

For this PR I'm fine with this but I think we should definitely do it more static and type checked in the future.

src/frequenz/client/dispatch/_types.py
```diff
@@ -0,0 +1,442 @@
...
+        pb_request.is_active = self.is_active
+        pb_request.is_dry_run = self.is_dry_run
+        pb_request.payload.update(self.payload)
+        pb_request.recurrence.CopyFrom(self.recurrence.to_protobuf())
+
+        return pb_request
+
+
+@dataclass(kw_only=True)
+class Dispatch(DispatchCreateRequest):
```
llucax: I wouldn't do this, besides both could diverge, this means someone could accidentally pass a `Dispatch` object where a `DispatchCreateRequest` is expected. I know it should be unlikely because `DispatchCreateRequest` should only be used internally, but that's actually another issue, we should better separate classes that are supposed to be public from classes that should be private, and public classes shouldn't inherit from private classes as users shouldn't have access to the private class.

src/frequenz/client/dispatch/_types.py
```diff
@@ -0,0 +1,442 @@
...
+
+        Returns:
+            The initialized dispatch.
+        """
+        return Dispatch(
+            id=_id,
+            create_time=create_time,
+            update_time=update_time,
+            **_request.__dict__,
+        )
```
llucax: This is not even used, right?

llucax (commented):
Besides the minor (optional) comments, I'm mostly OK to merge this as is, but there are a few more *major* points that I think we should change in the future:

* Make `list` async iterable
* Make `update` type safe
* Separate public from private classes more clearly (I would put them in different files)
* Make sure public classes don't inherit from private classes

PS: I didn't had a look at the tests yet

src/frequenz/client/dispatch/_client.py
Marenz: >  the code is quite convoluted with the huge match statement

It would look similar with any other solution as I have to take care / convert every (type of ) member individually.. if it's not a `match` statement it would be a `if para: ...` instead for each parameter..

src/frequenz/client/dispatch/_types.py
Marenz: for tests mostly

src/frequenz/client/dispatch/_types.py
Marenz: I really hate the repetition that that will cause...

src/frequenz/client/dispatch/_client.py
llucax: Yeah, is true that you will still check which ones are set to be updated and which are not one by one, but you could at least group `RecurrenceRuleUpdate` in wrapper, so that can be omitted as a group if nothing has to be changed for example, but the main point here for me is still type safety and being able to get errors at *`mypy`time* :D

llucax: I though about other options, using `TypedDict` for example, to avoid the bloat, but all have their issues, I guess this is something we might need to think more about, so I would leave it as in this PR and create an issue about it. I think how we do updates should also be consistent among all API clients, so is something we might want to discuss with the other API teams.

src/frequenz/client/dispatch/_types.py
llucax: Why not adding it to the `tests/` folder then? I you need to reuse it from different files you can just add something like `tests/utils.py`.

src/frequenz/client/dispatch/_client.py
Marenz: `ComponentSelector` itself is `ComponentSelector = list[int] | ComponentCategory` should that also become an Iterator?

src/frequenz/client/dispatch/_client.py
Marenz: > But it should be also possible to always create the PB object with all Nones and it will have the same effect?

I'd need to research that to be sure, right now, no idea

src/frequenz/client/dispatch/_client.py
Marenz: This is the create function, it doesn't return anything :)

src/frequenz/client/dispatch/_client.py
Marenz: A data class with every type optional as a parameter could be a way.

src/frequenz/client/dispatch/_types.py
llucax: Again, this is supposed to be stable, is a one time thing.

src/frequenz/client/dispatch/_client.py
llucax: Yeah, we would need to use a sentinel to indicate if a value was set or not (because you might need to update something from a value to `None`, so `None` can't mean "this attribute is not updated") but that still requires you to check one by one every attribute to see if it was overridden or not. `dict`s are the only ones that have the nice property of encoding the absence of a field natively, so you can just iterate over the existing keys to set the attributes in the message.

src/frequenz/client/dispatch/_client.py
llucax: you mean the `list[int]`? Nah, I think that should still be a list, because otherwise it would be more error prone when using as an attribute, if people iterate over the attribute without realizing they "consumed" the iterator.

In this case is safe to take an iterator as argument because we know we only need to iterate it and once. The downside of making it an iterator is if at some point we need to do more we either need to make a list of the iterator ourselves, or change the interface to take a list. But I guess this is unlikely to happen.

src/frequenz/client/dispatch/_client.py
Marenz: https://github.com/frequenz-floss/frequenz-client-dispatch-python/issues/15

src/frequenz/client/dispatch/_client.py
Marenz: I could also use the top suggestion and use `_` instead of `.` as path separator..

leandro-lucarella-frequenz: For what exactly?

src/frequenz/client/dispatch/_client.py
leandro-lucarella-frequenz: I would still like to confirm users can't create a new dispatch without a `ComponentSelector`. @stefan-brus-frequenz

leandro-lucarella-frequenz (approved):

src/frequenz/client/dispatch/_client.py
Marenz: For making `**new_fields: Any` possible

PR #21, opened 2024-02-29
Make test classes available

Marenz:
- Move conftest outside of dispatch namespace
- Move Dispatch.id attribute to the top
- Update dependency to use to/from datetime functions
- ~Also export the basic type `Dispatch`~
- Move MockServices, MockClient and Sampler to `test` package


src/frequenz/client/conftest.py
llucax, line 1: Why this change?

src/frequenz/client/dispatch/test/__init__.py
```diff
@@ -0,0 +1,10 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Helpful utilities for testing with the dispatch client."""
+
+from .sampler import Sampler
+from .service import MockClient as Client
```
llucax: I wouldn't rename this, I think it is still more clear to keep the `Mock` prefix even if it is in the `test` package.

src/frequenz/client/dispatch/test/sampler.py
```diff
@@ -0,0 +1,115 @@
...
+from frequenz.client.dispatch.types import (
+    Dispatch,
+    EndCriteria,
+    Frequency,
+    RecurrenceRule,
+    Weekday,
+)
+
+
+class Sampler:
```
llucax: If this is going to be public, I would use more clear names, like `DispatchGenerator`.

src/frequenz/client/dispatch/test/sampler.py
```diff
@@ -0,0 +1,115 @@
...
+            ],
+            bymonthdays=[
+                self._rng.randint(1, 31) for _ in range(self._rng.randint(0, 10))
+            ],
+            bymonths=[
+                self._rng.randint(1, 12) for _ in range(self._rng.randint(0, 12))
+            ],
+        )
+
+    def __call__(self) -> Dispatch:
```
llucax:
```suggestion
    def generate_dispatch(self) -> Dispatch:
```
?

src/frequenz/client/dispatch/test/sampler.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, seed: int = 0) -> None:
+        """Initialize the sampler.
+
+        Args:
+            seed: seed to initialize the rng with
+        """
+        self._rng = random.Random(seed)
+        self._last_id: int = 0
+
+    def recurrence_rule(self) -> RecurrenceRule:
```
llucax:
```suggestion
    def generate_recurrence_rule(self) -> RecurrenceRule:
```

src/frequenz/client/dispatch/test/sampler.py
llucax, line 1: Probably not for this PR, as you are just moving existing code around, but have you considered creating a hypothesis strategy instead? Or maybe both, so we don't force users to use hypothesis.

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,259 @@
...
+# pylint: enable=no-name-in-module
+from frequenz.client.dispatch import Client
+from frequenz.client.dispatch._internal_types import DispatchCreateRequest
+from frequenz.client.dispatch.types import Dispatch
+
+__all__ = ["Service", "MockClient"]
+
+
+@dataclass
+class Service:
```
llucax:
```suggestion
class MockService:
```

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,259 @@
...
+
+    dispatches: list[Dispatch]
+
+    last_update_request: DispatchUpdateRequest | None = None
+
+    _last_id: int = 0
+
+    def __init__(self) -> None:
+        """Initialize the mock service."""
+        self.dispatches = []
```
llucax: Are defaults properly initialized even if you write your own `__init__`? Any reason not to do this? Which allow initialization passing a list of dispatches already?

```suggestion
    dispatches: list[Dispatch] = dataclasses.field(default_factory=list)

    last_update_request: DispatchUpdateRequest | None = None

    _last_id: int = 0
```

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,259 @@
...
+__all__ = ["Service", "MockClient"]
+
+
+@dataclass
+class Service:
+    """Dispatch mock service for testing."""
+
+    dispatches: list[Dispatch]
+
+    last_update_request: DispatchUpdateRequest | None = None
```
llucax: These should have documentation if they are public.

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,259 @@
...
+    # pylint: enable=invalid-name
+
+
+class MockClient(Client):
+    """Mock client for testing."""
+
+    def __init__(self) -> None:
+        """Initialize the mock client."""
+        super().__init__(grpc.aio.insecure_channel("mock"), "mock")
+        self._stub = Service()  # type: ignore
```
llucax: So, if I understand this correctly, users to add fake dispatches to the mock service/client need to do something like:

```py
mock_client._stub.dispatches.append()
```
?

If so, I think you should provide some `add_dispatch()` and `remove_dispatch()` or so to the mock client, or a R/W `@property` to access the underlying `dispatches` list as `mock_client.dispatches` directly.

llucax (commented):
In general I like this very much, and was planning to go on a similar direction for the microgrid API.

I'm a bit concerned about trying to sort of simulate the server instead of mocking it (for example, implementing the filtering logic), as it could get out of hand as the server becomes more complex. We shouldn't encourage to write unit tests that involves so high level of "simulation", but just mocks whatever it is using at the dumbest level possible. So instead I would probably just purely mock the return of the function. But I guess that is a broader discussion to have elsewhere.

I would also like to [register a `pytest` fixture[(https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#making-your-plugin-installable-by-others) so users can just write tests like:

```py
def test_my_app(microgrid_client: MockMicrogridClient) -> None:
    micorgrid_client.mock_whatever()
```

Like we do with `time_machine` and other plugins.

src/frequenz/client/dispatch/test/service.py
Marenz: No, it's way simpler. You can use the client directly, just the normal create/list/update methods.
All the user needs to do is create an instance of `mock_client` really and can use the methods directly.

Marenz:
> I'm a bit concerned about trying to sort of simulate the server instead of mocking

I generally agree with that concern, but as long as the implementation is bellow 300 lines, I felt it was a non-issue.

src/frequenz/client/dispatch/test/service.py
llucax: Oh, ok, I see. For me this is not a mock at all and I wouldn't call it that. A mock in general is super shallow, has almost no logic except for registering which stuff you used from it so you can then assert that what you expected to happen really happened. I think this should be called a fake or simulation. And I wouldn't encourage it's use for **unit tests**.

> Fakes are objects that have working implementations. On the other hand, mocks are objects that have predefined behavior.

https://medium.com/@mohammad.roshandelpoor/whats-the-difference-between-faking-mocking-and-stubbing-2f5aab410531

Seems really cool for pseudo-integration tests though (as is not a pure integration test either, as it's not the real server).

But this also makes me think why not using the real server for this instead.

Again, probably a bigger topic affecting all the APIs.

Marenz: Well, they are very useful for unit tests, it's easier to construct the data for the expected response this way instead of mocking up the exact response each time.. but I can change the code to do that instead. 

Note that this is exactly what we talked about already in slack, I specifically asked about exactly this ;)

llucax: Haha, yeah, but sometimes you don't realize until you see the idea in the largest context. Also it is different when something is internal than when you make it public though.

So, for narrowing things down to the context of this PR, I would just rename `MockClient` to `FakeClient` and not expose the `Service` class at all, users of the clients don't need to know about protobuf stuff.

About:
> Well, they are very useful for unit tests, it's easier to construct the data for the expected response this way instead of mocking up the exact response each time.. but I can change the code to do that instead.

I understand it might be more convenient but I think is the wrong approach **for unit tests**. Why should a **unit test** for listing dispatches should fail if the create method of the client is broken?

Marenz: So you're concerned about the missing separation, that's fair. Though I do think I do access the internals of the FakeServer for that check in test_create ;)

tests/test_dispatch_client.py
```diff
@@ -3,366 +3,15 @@
...
-
-
-class MockClient(Client):
-    """Mock client for testing."""
-
-    def __init__(self) -> None:
-        """Initialize the mock client."""
-        super().__init__(grpc.aio.insecure_channel("mock"), "mock")
-        self._stub = DispatchMockService()  # type: ignore
+# pylint: disable=no-name-in-module
```
llucax: Why is this needed?

src/frequenz/client/dispatch/test/service.py
llucax: I didn't get that.

llucax: BTW and FWIW, I don't think I'm being inconsistent with what I said on slack:

> I can't make an informed opinion, I feel I need getting my hands dirty first too to judge. My gut feeling says that it is probably wise to start with less abstraction (just mocking every function in every test) but at some point we'll find patterns and will want to abstract it away.
> But usually from the unit test point of view, I think when you mock in the test itself it is much more obvious what you are testing exactly, which is useful, so at some point is a matter of better readability and more repetition or the other way around.

src/frequenz/client/conftest.py
Marenz: I thought it's nicer if this infrastructure file isn't mixed with the src files

tests/test_dispatch_client.py
Marenz: remainder from old code, will remove

src/frequenz/client/conftest.py
llucax: We put it in `src/` before, but I moved it because it makes harder to navigate files in GitHub (you won't get a top-level direct link to where the actual code lives but where the `conftest.py` lives. Also if you regenerate the templates the file will be put again in the "leaf" directory. I think both sucks and I wish we can get rid of the `conftest.py` hack soonish. I'm fine either way, just wanted to let you know the tradeoffs.

src/frequenz/client/dispatch/test/service.py
Marenz: > I understand it might be more convenient but I think is the wrong approach for unit tests. Why should a unit test for listing dispatches should fail if the create method of the client is broken?

Alright, true, but this argument only counts for the tests of the client itself
Elegantely, I dodged the scenario you described by [combining](https://github.com/frequenz-floss/frequenz-client-dispatch-python/blob/v0.x.x/tests/test_dispatch_client.py#L383) those specific tests XD

Though the whole fake server is still useful for testing things that build on top of the client like the dispatch actor.

Marenz:
* Named client/service `Fake*`
* Made service available through client `.service`
* Used that to add new tests that directly set or get the expected dispatches

pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
   "frequenz-api-dispatch @ git+https://github.com/frequenz-floss/frequenz-api-dispatch.git@223315c",
 #  "frequenz-api-dispatch >= 0.13.0, < 0.14",
   # Directly use unreleased commit until the next release
-  "frequenz-client-base @ git+https://github.com/frequenz-floss/frequenz-client-base-python.git@57d3093",
+  "frequenz-client-base @ git+https://github.com/frequenz-floss/frequenz-client-base-python.git@03f2ed2",
```
llucax: There is a release available now, so you could use:
```suggestion
  "frequenz-client-base >= 0.2.1, < 0.3.0",
```

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,264 @@
...
+# pylint: enable=no-name-in-module
+from frequenz.client.dispatch import Client
+from frequenz.client.dispatch._internal_types import DispatchCreateRequest
+from frequenz.client.dispatch.types import Dispatch
+
+__all__ = ["FakeClient"]
+
+
+@dataclass
+class FakeService:
```
llucax: You might want to update the commit message too, it still says `MockXxx`.

src/frequenz/client/dispatch/test/__init__.py
```diff
@@ -0,0 +1,9 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Helpful utilities for testing with the dispatch client."""
+
+from .generator import DispatchGenerator, to_create_params
+from .service import FakeClient, FakeService
+
+__all__ = ["DispatchGenerator", "FakeClient", "FakeService", "to_create_params"]
```
llucax: I wouldn't do this for public modules, it is confusing to be able to find the same symbol in multiple places. I'd just make the sub-modules private (start with `_`) or leave this `__init__` empty. The advantage or having stuff more scattered is that you don't need to import unnecessary stuff if you want to only use, say, the `DispatchGenerator`. The disadvantage is that you need to add more imports if you want to use the whole thing. I guess I would go with the later approach for simplicity, because it should be pretty unlikely that you want to use one but not the other.

src/frequenz/client/dispatch/__init__.py
```diff
@@ -4,5 +4,6 @@
 """Dispatch API client for Python."""
 
 from ._client import Client
+from .types import Dispatch
 
-__all__ = ["Client"]
+__all__ = ["Client", "Dispatch"]
```
llucax: Same as with `test/__init__.py` here, I would make the `types` module private (`_types`), as it is very unlikely that anyone would like to use the `Dispatch` type without using the client.

src/frequenz/client/dispatch/test/service.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self) -> None:
+        """Initialize the mock client."""
+        super().__init__(grpc.aio.insecure_channel("mock"), "mock")
+        self._stub = FakeService()  # type: ignore
+
+    @property
+    def service(self) -> FakeService:
+        """Get the fake service."""
+        return self._stub  # type: ignore
```
llucax: I would move this to `client.py` and make `service.py` private (`_service.py`), so you can use it in the tests for the client itself (in this repo) but it is still hidden to users of the client, which shouldn't do anything related to protobuf.

And if possible (and not very annoying for the tests of this repo), I would not make the `service` property public, if you are doing this only to provide access to the `service.dispatches` list, then add a `dispatches` property that returns `self._stub.dispatches`, assuming these dispatches are the wrappers and not the protobuf types.

llucax: OK, I see now that you are using `__all__` to *hide* the `FakeService`. I'm not sure if this will hide it from the generated docs too (you can try), but still I would recommend going for the splitting approach for 2 reasons:

1. Consistency, other projects are using that approach.
2. The user of this library will import `from ...test.service import FakeClient`. It is a bit weird to have `service` in there.

src/frequenz/client/dispatch/test/generator.py
```diff
@@ -0,0 +1,138 @@
...
+        "microgrid_id": dispatch.microgrid_id,
+        "_type": dispatch.type,
+        "start_time": dispatch.start_time,
+        "duration": dispatch.duration,
+        "selector": dispatch.selector,
+        "active": dispatch.active,
+        "dry_run": dispatch.dry_run,
+        "payload": dispatch.payload,
+        "recurrence": dispatch.recurrence,
+    }
```
llucax: Minor, but I would move this to the `_client.py` module (considering you move the `FakeClient` there, as this is more releated to the client than to dispatches generation.

src/frequenz/client/dispatch/test/service.py
llucax: I recommend using a `typing.cast` here to make the intention more clear and avoid masking other type errors.
```suggestion
        return cast(FakeService, self._stub)
```

tests/test_dispatch_client.py
```diff
@@ -3,393 +3,56 @@
...
+
+async def test_list_dispatches() -> None:
+    """Test listing dispatches."""
+    sampler = DispatchGenerator()
+    client = FakeClient()
+
+    client.service.dispatches = [sampler.generate_dispatch() for _ in range(100)]
+
+    async for dispatch in client.list(microgrid_id=1):
+        assert dispatch in client.service.dispatches
```
llucax: Nice, I like this (even better with `client.dispatches` instead).

llucax (commented):
Just a few last comments, mainly about modules structure.

Also you have a trailing dot in the commit updating release notes. Please train your ChatGPT accordingly :smile:

src/frequenz/client/dispatch/__init__.py
Marenz: Hmm in this case, `Dispatch` is a very common type, but there are quite a few sub-types that are def. useful (Weekday Enum, Frequency) but not always required when just `Dispatch` is required

llucax: Then just leave it in `types` and don't expose it here, symbols should be exposed only in one place. And also there is no benefit, with this, you are actually forcing the user to import `types`, even if they don't use **any** type from there, every time they do `import frequenz.client.dispatch`.

We have been switching between a more nested and a flatter structure for a while now, I'm not convinced which is best, but when it is very likely that you'll use many stuff and the 'submodules' are relatively small, I think the flatter structure makes things easier. I think that is the case for the client, so I would probably go with that, but if you prefer to keep `types` separate, OK, just don't mix both styles and make sure symbols are publicly exposed only in one place.

src/frequenz/client/dispatch/test/service.py
Marenz: You noticed that, eh? :eyes:

src/frequenz/client/dispatch/test/__init__.py
Marenz: I am a bit confused for which approach you're arguing now. 
The arguments you say sound to me like "keep commonly used things together" no?

llucax: Together how? (same file? Same module/package?) :laughing: 

But no, my main argument is against exposing symbols publicly in different locations and being consistent. The rest have different trade-offs and might be valid.

llucax: Specifically in this case I'm suggesting to make all modules in `test` private: `test/_service.py`, `test/_generator.py`, etc. and then keep the `test/__init__.py as it is now, except for `FakeService` that I think it shouldn't be added because **downstream users** of this project don't need it and shouldn't deal with protobuf stuff.

llucax: Then a separate issue is if you want to split those private modules (`test/_service.py`, etc.) into more or less files. Under this scheme, where the sub-module structure is just an implementation detail, it's not all that important how you split the stuff, as it is not part of the public interface. I would still try to keep modules small so they are manageable, and to groups very closely related stuff in the same module (specially stuff that needs to know about the implementation details of other stuff or access private data).

Marenz: Ah, I went the other way now and removed all from init

llucax: OK, I think the repo is still having some mixed style, for example now we have stuff in `types.py` that shouldn't be really public (the functions to convert to/from protobuf, but we are also exposing this as methods, so it is a different refactor that we need there, and something more for the future as we need to coordinate with other APIs). Long term I think it would be simpler to have a flat structure for the client, so users don't need to import stuff from different nested modules. In terms of *performance* or dependencies, when you import `__init__` you are importing `_client` which is importing `types` anyway, so is not like you are saving anything, and I guess it would be pretty unlikely that users import only the types, but maybe I'm wrong.

Anyway, out of scope for this PR, so approving.

llucax (approved):

PR #26, opened 2024-03-07
Useful additions for testing

Marenz:
- Add and use `DummyChannel` instead of grpc one
- Generator: Allow specifying a microgrid id
- Move pylint warning switch down

### DIFF
--- a/src/frequenz/client/dispatch/test/_service.py
+++ b/src/frequenz/client/dispatch/test/_service.py
@@ -123,12 +123,15 @@ class FakeService:
         request: DispatchUpdateRequest,
     ) -> Empty:
         """Update a dispatch."""
-        dispatch = next((d for d in self.dispatches if d.id == request.id), None)
+        index = next(
+            (i for i, d in enumerate(self.dispatches) if d.id == request.id),
+            None,
+        )
 
-        if dispatch is None:
+        if index is None:
             return Empty()
 
-        pb_dispatch = dispatch.to_protobuf()
+        pb_dispatch = self.dispatches[index].to_protobuf()
 
         # Go through the paths in the update mask and update the dispatch
         for path in request.update_mask.paths:
@@ -173,6 +176,8 @@ class FakeService:
 
         dispatch = Dispatch.from_protobuf(pb_dispatch)
         dispatch.update_time = datetime.now(tz=timezone.utc)
+
+        self.dispatches[index] = dispatch
 
         return Empty()
 
--- a/src/frequenz/client/dispatch/test/_service.py
+++ b/src/frequenz/client/dispatch/test/_service.py
@@ -5,7 +5,6 @@
 
 Useful for testing.
 """
-
 import dataclasses
 from dataclasses import dataclass
 from datetime import datetime, timezone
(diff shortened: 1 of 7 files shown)


llucax (commented):
Wait, why not `mock_channel = MagicMock(spec=grpc.insecure_channel)` (or the base class used by all grpc channels)?

If there was a use for a proper dummy channel then it's probably better to put it in `frequenz-client-base`.

Marenz:
> Wait, why not mock_channel = MagicMock(spec=grpc.insecure_channel) (or the base class used by all grpc channels)?
>
> If there was a use for a proper dummy channel then it's probably better to put it in frequenz-client-base.

Your example fails for me:

```
    def __getattr__(self, name):
        if name in {'_mock_methods', '_mock_unsafe'}:
            raise AttributeError(name)
        elif self._mock_methods is not None:
            if name not in self._mock_methods or name in _all_magics:
>               raise AttributeError("Mock object has no attribute %r" % name)
E               AttributeError: Mock object has no attribute 'unary_unary'

/usr/lib64/python3.11/unittest/mock.py:653: AttributeError
```

But it turns out all I needed is `MagicMock()` and it worked. So I don't need the dummy channel after all.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -29,9 +28,8 @@
 from frequenz.api.dispatch.v1.dispatch_pb2 import DispatchUpdateRequest
 from google.protobuf.empty_pb2 import Empty
 
-from frequenz.client.base.conversion import to_datetime as _to_dt
-
 # pylint: enable=no-name-in-module
+from frequenz.client.base.conversion import to_datetime as _to_dt
 from frequenz.client.dispatch._internal_types import DispatchCreateRequest
 from frequenz.client.dispatch.types import Dispatch
```
shsms: maybe relative imports are better here?

shsms (approved):
Just a comment about code that's not part of this PR.

Regarding the use of MagicMock to mock a grpc connection, that looks very interesting.  if you have some working tests, maybe you could give a demo during the next api meeting.

Marenz:
added commit to make imports relative

shsms (approved):
I guess you need a second approval, but looks good to me.

stefan-brus-frequenz (approved):

llucax:
> Your example fails for me:

Oh, that's probably because of how the extremely dynamic Google's grpc is (`spec=...` use introspection to also check that only symbols that exist in the mocked object are called, and with the right arguments). Another reason to use betterproto, google's stuff for python really sucks...

src/frequenz/client/dispatch/types.py
```diff
@@ -148,7 +148,7 @@ def to_protobuf(self) -> PBRecurrenceRule.EndCriteria:
 
 
 # pylint: disable=too-many-instance-attributes
-@dataclass(frozen=True, kw_only=True)
+@dataclass(kw_only=True)
```
llucax, line 151: Having immutable classes is useful. I don't remember by heart now how complex or simple this one is, but with dataclasses is always easy to create a new instance from another by [`replace()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) function.

Not saying this is the right approach, just saying that there is a simple alternative to update immutable dataclasses.

Marenz: Dispatches are objects meant to be manipulated though, (except for the ones that execute it at the end of course). So most use cases become easier when you can modify it.

llucax: Yeah, true.

llucax: Well, actually not entirely.

I think the normal case is you create one, for which you don't need to mutate an existing one, and then you *consume* it, where you probably don't want to modify them either. For these 2 cases immutability seems like a nice safety feature.

Of course you could also update one, in that case using replace doesn't sound too cumbersome, and it is probably not a very common scenario, you probably made a mistake when creating it.

So IMHO having immutable `Dispatch` object might still be the best approach.

Marenz: Fair, I can live with immutable dispatch

PR #32, opened 2024-04-11
Make `Dispatch` class frozen

Marenz:
Signed-off-by: Mathias L. Baumann <mathias.baumann@frequenz.com>


src/frequenz/client/dispatch/test/_service.py
```diff
@@ -175,7 +175,10 @@ async def UpdateMicrogridDispatch(
                             )[:]
 
         dispatch = Dispatch.from_protobuf(pb_dispatch)
```
llucax, line 177: Another approach here might be to add an option here.

```py
def from_protobuf(cls, pb_object: PBDispatch, *, set_update_time: bool = True) -> Dispatch:
    ...
```

But then I don't understand why you change the update time that comes from the API.

llucax (commented):
Just one question about the update time.

src/frequenz/client/dispatch/test/_service.py
Marenz: > why you change the update time that comes from the API.

Because we're in the `test` module ;)

llucax: Ah, I missed that detail, then `replace()` is perfectly fine, sorry for the noise!

llucax (approved):

PR #33, opened 2024-04-16
Interface fixes & enhancements

Marenz:
- **Fix start_time conversion imprecision bug**
- **Rename client-side create request fields `is_{active,dry_run}`**
- **Make `.create()` return the new created dispatch**


RELEASE_NOTES.md
```diff
@@ -12,6 +12,7 @@
 
 - A new module `frequenz.client.dispatch.test` has been added, providing a fake Service and Client as well as a `DispatchGenerator` to generate `Dispatch` instances filled with random data.
 - The `DispatchGenerator.generate_dispatch` method now accepts a `microgrid_id` parameter to generate `Dispatch` instances with a specific microgrid ID.
+- The `Client.create()` method now returns by default the newly created `Dispatch` object. This behavoir can be changed using the `fetch_resulting_dispatch` parameter.
```
llucax: If the whole module was added in this release, there is no need to document changes to unreleased code, I think having only the "A new module..." entry is enough for this release.

src/frequenz/client/dispatch/_client.py
```diff
@@ -255,3 +271,34 @@ async def delete(self, dispatch_id: int) -> None:
...
+                "type",
+                "start_time",
+                "duration",
+                "selector",
+                "active",
+                "dry_run",
+                "recurrence",
+            ]:
+                if getattr(request, attribute) != getattr(dispatch, attribute):
+                    return None
```
llucax: So we don't even get the ID when we create a dispatch :grimacing: ?

Is this only excluding the `update_time` or more stuff too? Maybe it is worth a comment.

src/frequenz/client/dispatch/_client.py
```diff
@@ -130,9 +130,16 @@ async def create(
         dry_run: bool = False,
         payload: dict[str, Any] | None = None,
         recurrence: RecurrenceRule | None = None,
-    ) -> None:
+        fetch_resulting_dispatch: bool = True,
+    ) -> None | Dispatch:
```
llucax: Any reason not to always fetch it for now and make it transparent when the server supports returning the newly created dispatch?

llucax (commented):
It might be useful to comment in more detail why there was any imprecision in "Fix start_time conversion imprecision bug" and how it is fixed by the common functions, at least for me it is not clear so I might make the same mistake in other API clients.

RELEASE_NOTES.md
Marenz: Hu, no, `Client.create()` is an existing core method

llucax: Oh, OK. But still there is an entry above documenting a change to the test module right? Unrelated to this PR though.

src/frequenz/client/dispatch/_client.py
Marenz: My only thought was that it takes another round-trip, that's why I made it optional opt-out for now. I can make it always fetch it, sure.

src/frequenz/client/dispatch/_client.py
Marenz: >  So we don't even get the ID when we create a dispatch 

Yeah, nothing so far unfortunately

> Is this only excluding the update_time or more stuff too? Maybe it is worth a comment.

Now that you mention it, it is basically not excluding anything, it's just including everything from the request type.
I changed the code accordingly...

RELEASE_NOTES.md
Marenz: Ah yes. I see what you mean. I suppose we'll fix that with the release then.

Marenz:
Updated:
* always return Dispatch in create now (no parameter
* use __dict__ of CreateRequest for attribute comparison
* Added explanation for time bug

src/frequenz/client/dispatch/_client.py
llucax: I guess we won't be using the create much for now, so it might be good to get the interface as it should be in the future now if the impact is almost zero.

src/frequenz/client/dispatch/_client.py
```diff
@@ -255,3 +267,28 @@ async def delete(self, dispatch_id: int) -> None:
...
+
+        Args:
+            request: The dispatch create request.
+
+        Returns:
+            Dispatch: The created dispatch, if it was found.
+        """
+        async for dispatch in self.list(microgrid_id=request.microgrid_id):
+            found = True
+            for key, value in request.__dict__.items():
```
llucax, line 287: Are you sure this works as expected? When I look at `__dict__` in a message it also have `__slots__` and other stuff, maybe it would be better to use the [`google.protobuf.json_format.MessageToDict`](https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html#google.protobuf.json_format.MessageToDict) function?

You'll need to set at lest `including_default_value_fields=True` and I guess it would be better to set `use_integers_for_enums=True` and maybe something like `float_precision=8`.

leandro-lucarella-frequenz:
Oh dear, it seems like I lost a in-process review. I will try to recreate what I remember. I suggest opening a new PR with the new commits, as they are not really necessary to be in this PR and we were close to finish the review. I have a few comments for the new commits and that will make merging this Pr more time consuming.

Oh, damn, I was logged with the wrong user to do some admin tasks... :facepalm:

llucax:
I suggest opening a new PR with the new commits, as they are not really necessary to be in this PR and we were close to finish the review. I have a few comments for the new commits and that will make merging this PR more time consuming.

llucax:
> Added explanation for time bug

Awesome, thanks! Now I remember.

llucax (commented):
With this small change, this should be ready to merge up to the commit changing the `Client.create()` method.

src/frequenz/client/dispatch/_client.py
Marenz: It at least seems to be what I expected it to be..
These are the fields and their values I get in my loop
```
request.microgrid_id = 1
request.type = ECHO_FREQUENCY
request.start_time = 2024-04-18 18:44:42.901963+00:00
request.duration = 0:05:00
request.selector = ComponentCategory.INVERTER
request.active = True
request.dry_run = False
request.payload = {'font': 'Times New Roman'}
request.recurrence = RecurrenceRule(frequency=<Frequency.UNSPECIFIED: 0>, interval=0, end_criteria=None, byminutes=[], byhours=[], byweekdays=[], bymonthdays=[], bymonths=[])
```

Marenz: Ah, besides, this isn't a Protobuf object, this is our abstraction. We only convert to protobuf right before sending the request.

Marenz:
Responded to comment and removed the unrelated commits

src/frequenz/client/dispatch/_client.py
```diff
@@ -123,9 +123,12 @@ async def create(
         dry_run: bool = False,
         payload: dict[str, Any] | None = None,
         recurrence: RecurrenceRule | None = None,
-    ) -> None:
+    ) -> Dispatch:
         """Create a dispatch.
 
+        Will try to return the created dispatch, identifying it by
+        the same fields as the request.
```
shsms, line 130: Maybe present tense is nicer, but you could also elaborate a bit on what "try" means:

```suggestion
        Returns the created dispatch, if it can be identified using
        the same fields as the request.  If not, raises an exception.
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -255,3 +267,31 @@ async def delete(self, dispatch_id: int) -> None:
...
+        Returns:
+            Dispatch: The created dispatch, if it was found.
+        """
+        async for dispatch in self.list(microgrid_id=request.microgrid_id):
+            found = True
+            for key, value in request.__dict__.items():
+                if key == "start_time":
+                    value = rounded_start_time(value)
+
+                if not (found := getattr(dispatch, key) == value):
```
shsms, line 291: Since python 3.11 `getattr` can be much slower because they get in the way of the adaptive interpreter and other optimizations.  It is too bad `api.CreateDispatch` doesn't return a dispatch ID, or the entire dispatch.

Won't be a bad idea to implement this in the api level, I think, so that you can avoid additional calls to list and then filter.

shsms (commented):
Just a couple of comments - no immediate action needed for one of them I think.

src/frequenz/client/dispatch/_client.py
Marenz: > Won't be a bad idea to implement this in the api level, I think, so that you can avoid additional calls to list and then filter.

That's all in the works, this is just to provide the feature before the API is ready. That way we can later add it and clients don't need to change.

src/frequenz/client/dispatch/_client.py
Marenz: That's a fair suggestion, though given that we'll update this part soon, I think we can leave it for now.

ktickner (dismissed):

shsms:
Ok, LGTM then.

ktickner (approved):

ktickner (dismissed):

ktickner (dismissed):

Marenz:
Re-opeening as github seems very bugged now.

shsms:
Probably needs a rebase?

PR #34, opened 2024-04-18
Make `Client.update` require named parameters

Marenz:
I noticed it happened often that I used the microgrid id instead of
the dispatch id to identify the to-be-updated dispatch which of course
is wrong.

This should make it more obvious when using the API what kind of ID should
be passed.

Optimally we have distinct types for Microgrid ID and Dispatch ID though.


stefan-brus-frequenz (approved):
LGTM

PR #44, opened 2024-05-03
Don't recursively scan for pyproject.toml when calculating the cache key

llucax:
Don't recursively scan for pyproject.toml when calculating the cache key
    
We don't really need to scan for `pyproject.toml` files recursively, since we only have one in the root of the repository. This should make the cache key calculation more efficient and less error prone, as when using qemu, there are some files that are not accessible and the hash calculation fails.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -159,7 +159,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /tmp/pip-cache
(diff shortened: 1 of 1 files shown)


llucax:
Mmm, maybe putting the SDK team as an owner wasn't the best approach, as it now asks for a review to the SDK team too, which might be too much. What do you think? Ideally I would like that SDK team members can approve but are not asked for a review automatically.

daniel-zullo-frequenz:
>  I would like that SDK team members can approve but are not asked for a review automatically.

I agree that'd be the best

daniel-zullo-frequenz (approved):

PR #48, opened 2024-05-27
Support payload and recurrence flags in CLI/REPL

Marenz:
- **Remove commented dead code**
- **Add short-notation for some options**
- **Actually implement the CLI "payload" option**
- **Implement recurrence in the CLI/REPL**

### DIFF
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -102,8 +102,8 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
 @click.argument("start-time", required=True, type=FuzzyDateTime())
 @click.argument("duration", required=True, type=FuzzyTimeDelta())
 @click.argument("selector", required=True, type=SelectorParamType())
-@click.option("--active", type=bool, default=True)
-@click.option("--dry-run", type=bool, default=False)
+@click.option("--active", "-a", type=bool, default=True)
+@click.option("--dry-run", "-d", type=bool, default=False)
 @click.option("--payload", type=str, help="JSON payload for the dispatch")
 @click.option("--recurrence", type=str, help="Recurrence rule (see documentation)")
 @click.pass_context
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -148,9 +148,6 @@ async def update(
         raise click.BadArgumentUsage("At least one field must be given to update.")
 
     try:
-        # if duration := new_fields.get("duration"):
-        #    new_fields.pop("duration")
-        #    new_fields["duration"] = timedelta(seconds=int(duration))
         await ctx.obj["client"].update(dispatch_id=dispatch_id, new_fields=new_fields)
         click.echo("Dispatch updated.")
     except grpc.RpcError as e:
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -17,7 +17,13 @@ from prompt_toolkit.history import FileHistory
 from prompt_toolkit.patch_stdout import patch_stdout
 from prompt_toolkit.shortcuts import CompleteStyle
 
-from ._cli_types import FuzzyDateTime, FuzzyIntRange, FuzzyTimeDelta, SelectorParamType
+from ._cli_types import (
+    FuzzyDateTime,
+    FuzzyIntRange,
+    FuzzyTimeDelta,
+    JsonParamType,
+    SelectorParamType,
+)
 from ._client import Client
 
 DEFAULT_DISPATCH_API_HOST = "88.99.25.81"
@@ -104,7 +110,9 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
 @click.argument("selector", required=True, type=SelectorParamType())
 @click.option("--active", "-a", type=bool, default=True)
 @click.option("--dry-run", "-d", type=bool, default=False)
-@click.option("--payload", type=str, help="JSON payload for the dispatch")
+@click.option(
+    "--payload", "-p", type=JsonParamType(), help="JSON payload for the dispatch"
+)
 @click.option("--recurrence", type=str, help="Recurrence rule (see documentation)")
 @click.pass_context
 async def create(
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -17,11 +17,18 @@ from prompt_toolkit.history import FileHistory
 from prompt_toolkit.patch_stdout import patch_stdout
 from prompt_toolkit.shortcuts import CompleteStyle
 
+from frequenz.client.dispatch.types import (
+    EndCriteria,
+    Frequency,
+    RecurrenceRule,
+    Weekday,
+)
+
 from ._cli_types import (
     FuzzyDateTime,
     FuzzyIntRange,
     FuzzyTimeDelta,
-    JsonParamType,
+    JsonDictParamType,
     SelectorParamType,
 )
 from ._client import Client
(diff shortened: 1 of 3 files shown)


Marenz:
no release notes as there is no release with the CLI client yet and we're just adding features noone knew are missing

src/frequenz/client/dispatch/__main__.py
```diff
@@ -17,7 +17,20 @@
...
+    Weekday,
+)
+
+from ._cli_types import (
+    FuzzyDateTime,
+    FuzzyIntRange,
+    FuzzyTimeDelta,
+    JsonParamType,
+    SelectorParamType,
+)
```
llucax: I missed this in the pr introducing the CLI, do you really want to expose it's types in the general client module? I think since it is a completely different thing it might make sense to expose it via a cli subpackage, no?

src/frequenz/client/dispatch/__main__.py
```diff
@@ -92,6 +105,55 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
...
+            if kwargs.get("count") or kwargs.get("until")
+            else None
+        ),
+        byminutes=list(kwargs.pop("by_minute", [])),
+        byhours=list(kwargs.pop("by_hour", [])),
+        byweekdays=[
+            Weekday[weekday.upper()] for weekday in kwargs.pop("by_weekday", [])
+        ],
+        bymonthdays=list(kwargs.pop("by_monthday", [])),
+    )
```
llucax, line 134: Would it make sense to want the user if there are any unused `kwargs` left?

src/frequenz/client/dispatch/__main__.py
```diff
@@ -92,6 +105,55 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
...
+        return value
+
+    until = param.name == "until" and value
+    count = param.name == "count" and value
+
+    if (
+        count
+        and ctx.params.get("until") is not None
+        or until
+        and ctx.params.get("count") is not None
```
llucax: If you are just checking if the keyword exists in the `dict` you can use `"until" in ctx.params`.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -92,6 +105,55 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
...
+    ):
+        raise click.BadArgumentUsage("Only count or until can be set, not both.")
+
+    if value and ctx.params.get("frequency") is None:
```
llucax, line 153: Or `not in`.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -120,8 +247,12 @@ async def create(
     SELECTOR is either one of the following: BATTERY, GRID, METER, INVERTER,
     EV_CHARGER, CHP or a list of component IDs separated by commas, e.g. "1,2,3".
     """
+    # Remove keys with `None` value
+    kwargs = {k: v for k, v in kwargs.items() if v}
```
llucax: ⚠️ **This removes `0`s and `""`s too!** ⚠️

src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -167,3 +168,36 @@ def convert(
                 param,
                 ctx,
             )
+
+
+class JsonParamType(click.ParamType):
```
llucax: Oh, interesting, there is no JSON Param built-in?

src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -167,3 +168,36 @@ def convert(
...
+    """Click parameter type for JSON strings."""
+
+    name = "json"
+
+    def convert(
+        self, value: Any, param: click.Parameter | None, ctx: click.Context | None
+    ) -> dict[str, Any]:
```
llucax, line 180: No arrays supported? If so, maybe call it `JsonDictParamType` instead? And why not just passing the whole string to `json.loads()`?

src/frequenz/client/dispatch/__main__.py
Marenz: No, it is expected that kwargs are left at this point, and those that are left are by definition the ones that `create()` wants to receive

src/frequenz/client/dispatch/__main__.py
Marenz: I am also checking if the member is not set to `None` when found. Normally all of them added to the dict, even when not passed on the cmdline.

src/frequenz/client/dispatch/_cli_types.py
Marenz: I _am_ passing the whole string to `json.loads()`, I just add the object `{}`'s if they are missing as a user-input convenience.

you mean as the direct type of `payload`? I didn't think that is supported and it wouldn't fit with the payload type either? https://github.com/frequenz-floss/frequenz-client-dispatch-python/blob/v0.x.x/src/frequenz/client/dispatch/types.py#L278

src/frequenz/client/dispatch/_cli_types.py
Marenz: Couldn't find it here: https://click.palletsprojects.com/en/8.1.x/api/#types

src/frequenz/client/dispatch/__main__.py
Marenz: At this point the `cli` subpackage would only contain one file (if we keep __main__ were it is).
I have no problem changing the location(s) though

src/frequenz/client/dispatch/__main__.py
Marenz: That is indeed a bug, will extend test-suite

src/frequenz/client/dispatch/__main__.py
llucax: I see, so you use `python -m frequenz.client.dispatch`, right? Yeah, I think I would make it `frequenz.client.dispatch.cli` to keep stuff separated, I don't think it is a good idea to pollute the top-level package used from other downstream projects with the CLI. Even more, eventually I would like to ship it as a separate `pip` package. So you could have a `src/frequenz/client/dispatch/cli/__init__.py`, `src/frequenz/client/dispatch/cli/_main.py` and `src/frequenz/client/dispatch/cli/_types.py` for example.

But for a separate PR:
* #49

src/frequenz/client/dispatch/__main__.py
llucax: OK, I see, and do you eventually warn if there are still kwargs left at some point when you finish popping stuff from it?

src/frequenz/client/dispatch/__main__.py
Marenz: can you have multiple packages per project? :thinking:

src/frequenz/client/dispatch/__main__.py
Marenz: It would cause an error in `create()` because those are then unexpected arguments

src/frequenz/client/dispatch/_cli_types.py
llucax: OK, I see, but if the return type is `dict` then you are allowing only `dict`s to be specified, can't you do something like `--json 1.0` or `--json "[1,2,3]"`, right? I guess specifically for `payload`, that is a `dict` indeed this is OK, I was jumping into the future and thinking about when it is moved to `client-base` or somewhere more general :smiling_imp:

Marenz: Oh, I misunderstood your phrasing. Because no  (top-level) arrays are supported it should be called diferently. Fair.

llucax (commented):
Made some extra comments, commenting on a review because I accidentally clicked on the re-request review to myself so you ask for another review when the PR is ready again.

src/frequenz/client/dispatch/__main__.py
llucax: Yes. You need to use multiple `pyproject.toml` files, it would be like a tiny mono-repo, but I guess since both would be so tied together it is fine. Another option would be to make it a separate repo but I guess that will be more overkill.

It will require some structural changes, like they should probably live in different parts of `src/`, like `src/frequenz/client/dispatch` and `src/frequenz/cli/dispatch` (so we can map one directory to each `pyproject.toml`), and might need some adjustments to repo-config, so definitely not in the short term, but in some future we might go for it.

src/frequenz/client/dispatch/_cli_types.py
llucax: Just in case you forgot to push or something, the name of this class didn't change. Or maybe I misunderstood and you didn't plan to change it?

llucax (dismissed):

src/frequenz/client/dispatch/_cli_types.py
Marenz: I forgot, yes. Fixed

llucax (dismissed):

Marenz:
@llucax merge queue tests failed, I pushed some fixes to make them more reliable

llucax (approved):

PR #55, opened 2024-06-03
Fix broken CI fix that prevented release

Marenz:
Fix broken CI fix that prevented release


llucax (approved):
I recommend doing the same for the `nox-all` job, although is not strictly necessary for this publishing issue, so approving anyway.

PR #56, opened 2024-06-13
Implement key-based authentication

Marenz:
- **Add key-based authentication support**
- **Update REPL client to work better with host/port/key parameters**

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* An API key for authorization must now be passed to the `DispatchClient`.
 
 ## New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,7 @@
 ## Upgrading
 
 * An API key for authorization must now be passed to the `DispatchClient`.
+* The client constructor now requires named parameters. If you are using the client directly, you will need to update your code to use named parameters.
 
 ## New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,7 +11,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* TLS is now enabled by default for the CLI client.
 
 ## Bug Fixes
 
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -37,18 +37,19 @@ DEFAULT_DISPATCH_API_HOST = "88.99.25.81"
 DEFAULT_DISPATCH_API_PORT = 50051
 
 
-def get_client(host: str, port: int) -> Client:
+def get_client(host: str, port: int, key: str) -> Client:
     """Get a new client instance.
 
     Args:
         host: The host of the dispatch service.
         port: The port of the dispatch service.
+        key: The API key for authentication.
 
     Returns:
         Client: A new client instance.
     """
     channel = grpc.aio.insecure_channel(f"{host}:{port}")
-    return Client(channel, f"{host}:{port}")
+    return Client(channel, f"{host}:{port}", key)
 
 
 # Click command groups
@@ -69,11 +70,17 @@ def get_client(host: str, port: int) -> Client:
     show_envvar=True,
     show_default=True,
 )
+@click.option(
+    "--key",
+    help="API key for authentication",
+    envvar="DISPATCH_API_KEY",
+    show_envvar=True,
+)
 @click.pass_context
-async def cli(ctx: click.Context, host: str, port: int) -> None:
+async def cli(ctx: click.Context, host: str, port: int, key: str) -> None:
     """Dispatch Service CLI."""
     ctx.ensure_object(dict)
-    ctx.obj["client"] = get_client(host, port)
+    ctx.obj["client"] = get_client(host, port, key)
 
 
 @cli.command("list")
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -5,7 +5,6 @@
 
 import asyncio
 import os
-import sys
 from pprint import pformat
 from typing import Any, List
 
@@ -53,7 +52,7 @@ def get_client(host: str, port: int, key: str) -> Client:
 
 
 # Click command groups
-@click.group()
+@click.group(invoke_without_command=True)
 @click.option(
     "--host",
     default=DEFAULT_DISPATCH_API_HOST,
@@ -75,12 +74,24 @@ def get_client(host: str, port: int, key: str) -> Client:
     help="API key for authentication",
     envvar="DISPATCH_API_KEY",
     show_envvar=True,
+    required=True,
 )
 @click.pass_context
 async def cli(ctx: click.Context, host: str, port: int, key: str) -> None:
     """Dispatch Service CLI."""
-    ctx.ensure_object(dict)
+    if ctx.obj is None:
+        ctx.obj = {}
+
     ctx.obj["client"] = get_client(host, port, key)
+    ctx.obj["params"] = {
+        "host": host,
+        "port": port,
+        "key": key,
+    }
+
+    # Check if a subcommand was given
+    if ctx.invoked_subcommand is None:
+        await interactive_mode(host, port, key)
 
 
 @cli.command("list")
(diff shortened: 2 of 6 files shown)


andrew-stevenson-frequenz:
I can't really comment on the Python, but is the GRPC connection encrypted? Or are we passing a symmetric key over a clear text channel?

Marenz:
> I can't really comment on the Python, but is the GRPC connection encrypted? Or are we passing a symmetric key over a clear text channel?

Same as reporting, currently clear text I think. 
At least some https will be added soon.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -37,22 +36,23 @@
 DEFAULT_DISPATCH_API_PORT = 50051
 
 
-def get_client(host: str, port: int) -> Client:
+def get_client(host: str, port: int, key: str) -> Client:
```
llucax: I suggest making these keywork-only for extra clarity now that we are taking 2 `str`, otherwise one could accidentally do a `get_client(key, port, host)` and maybe when that connection fails the `host` get logged, potentially leaking the key.

```suggestion
def get_client(*, host: str, port: int, key: str) -> Client:
```

src/frequenz/client/dispatch/_client.py
```diff
@@ -36,15 +36,17 @@
 class Client:
     """Dispatch API client."""
 
-    def __init__(self, grpc_channel: grpc.aio.Channel, svc_addr: str) -> None:
+    def __init__(self, grpc_channel: grpc.aio.Channel, svc_addr: str, key: str) -> None:
```
llucax: Same here:
```suggestion
    def __init__(self, *, grpc_channel: grpc.aio.Channel, svc_addr: str, key: str) -> None:
```

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -49,18 +49,56 @@ class FakeService:
...
+    def _check_access(self, metadata: grpc.aio.Metadata) -> None:
+        """Check if the access key is valid.
+
+        Args:
+            metadata: The metadata.
+
+        Raises:
+            grpc.RpcError: If the access key is invalid.
+        """
+        if metadata[0][0] != "key":  # type: ignore
```
llucax: A `cast` or a `isinstance()` runtime check might be better:

1. The `ignore` is more limited and doesn't mask other typing issues
2. The `cast`ed value keeps the correct type from now on
3. Intentions are much more clear
4. In the case of `isinstance` it actually checks that the type is correct, so for example a `AttributeError` can be properly converted to an auth error if the metadata is screwed.

I'm uncertain if this ignore is supposed to affect the first `[]`, the second, or what.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -49,18 +49,56 @@ class FakeService:
...
+            raise grpc.RpcError(
+                grpc.StatusCode.UNAUTHENTICATED,
+                "No access key provided",
+            )
+
+        key = metadata[0][1]  # type: ignore
```
llucax: Same.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -49,18 +49,56 @@ class FakeService:
...
+            raise grpc.RpcError(
+                grpc.StatusCode.PERMISSION_DENIED,
+                "Permission denied",
+            )
+
+        if key != "all":
+            raise grpc.RpcError(
+                grpc.StatusCode.UNAUTHENTICATED,
+                "Invalid access key",
+            )
```
llucax: Where are these `none` and `all` values coming from? It might be good to document it, and maybe even use `CONSTANT`s if they are specified somewhere.

src/frequenz/client/dispatch/test/_service.py
Marenz: > I'm uncertain if this ignore is supposed to affect the first [], the second, or what.

same

src/frequenz/client/dispatch/__main__.py
Marenz: For a local only singular-used once called function? Seems unnecessary to me..
Only to have a call like 

`get_client(host=host, port=port, key=key)` that really is not more informative than before ;)

llucax: It is a lot more informative! `get_client(host=host, port=port, key=key)` is verbose but obviously right, and `get_client(host=key, port=port, key=host)` is obviously wrong, while `get_client(key, port, host)` is not obviously wrong at all. This is the whole point.

If it is private I agree it is OK, although I see absolutely no reason why not to use the less error-prone form. Why do you like the error-prone form, only because it is more compact?

src/frequenz/client/dispatch/_client.py
```diff
@@ -166,7 +172,9 @@ async def create(
             recurrence=recurrence or RecurrenceRule(),
         )
 
-        await self._stub.CreateMicrogridDispatch(request.to_protobuf())  # type: ignore
+        await self._stub.CreateMicrogridDispatch(
+            request.to_protobuf(), metadata=self._metadata
+        )  # type: ignore
```
llucax, line 177: The `cast` comment applied to all `type: ignore`, I would also use a `cast(Awaitable[ResponseType], ...)` here and in all stub calls.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -35,6 +35,12 @@
 from .._internal_types import DispatchCreateRequest
 from ..types import Dispatch
 
+ALL_KEY = "all"
+"""Key that has access to all resources."""
+
+NONE_KEY = "none"
+"""Key that has no access to any resources."""
```
llucax: So, I guess this is only for development? I think it still needs some clarification, as it is it makes no sense, this is a public repo so everyone knows they can use "all" to have full authorization.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -49,18 +55,59 @@ class FakeService:
...
+    def _check_access(self, metadata: grpc.aio.Metadata) -> None:
+        """Check if the access key is valid.
+
+        Args:
+            metadata: The metadata.
+
+        Raises:
+            grpc.RpcError: If the access key is invalid.
+        """
+        # metadata is a weird tuple of tuples, we don't like it
```
llucax, line 67: > [!WARNING]
> A weird tuple of tuples is not the same as a dict.

A tuple of tuples is more like a multi-dict, where a key can have multiple values (the same happens with Mail/HTTP headers or URL query parameters). By just converting it to a dict blindly, you are getting a random value if a key has more than one value.

If you don't want to support multiple values for a key, I would get the last value for each key instead, I think it is more deterministic and probably more correct. Something like [`cgi.FieldStorage.getfirst`](https://docs.python.org/3/library/cgi.html#cgi.FieldStorage.getfirst) and [`cgi.FieldStorage.getlist`](https://docs.python.org/3/library/cgi.html#cgi.FieldStorage.getlist).

This will be probably needed by all APIs, so it might be worth putting into `client-base`.

src/frequenz/client/dispatch/__main__.py
Marenz: Well, yes. It was clear before. It was not really used anywhere except in this one instance, in that instance the case was very very clear, so I don't really see any advantage for this occasion but only costs to add something.. with no value (here)

src/frequenz/client/dispatch/_client.py
Marenz: That seems to be out of the scope of this PR though

src/frequenz/client/dispatch/test/_service.py
Marenz: defined in a file that creates a "fake service" created by a "fake client" in a package called "test" .. come on..

src/frequenz/client/dispatch/test/_service.py
Marenz: I doubt any client has to do this, given that this is the fake service...

Marenz: > By just converting it to a dict blindly, you are getting a random value if a key has more than one value.

I am not worried about that, I don't think we need or want to test that case in our tests.. (which is what this is for)

src/frequenz/client/dispatch/_client.py
llucax: True.

src/frequenz/client/dispatch/test/_service.py
llucax: chill

I didn't see the file name, sorry about that.

src/frequenz/client/dispatch/test/_service.py
llucax: Also missed it was only the test class. :innocent:

RELEASE_NOTES.md
```diff
@@ -10,7 +10,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* An API key for authorization can now be passed to the `DispatchClient`.
```
llucax: This might be better in `Upgrading` as it **must** be passed, right?

llucax (dismissed):

src/frequenz/client/dispatch/_client.py
Marenz: https://github.com/frequenz-floss/frequenz-client-dispatch-python/issues/57

llucax (approved):

PR #68, opened 2024-07-03
Ship our own root certificate until we have a proper one

Marenz:
To have any meaningful way to use the new TLS features, we're adding our own root certificate to validate the servers certificate until we get a proper one for the server from infrastructure.


hannah-stevenson-frequenz:
Seems a reasonable interim measure.

hannah-stevenson-frequenz (approved):

PR #73, opened 2024-08-09
Initial work to use base client

Marenz:
Currently on hold as the base client does not seem to easily support using a
custom root certificate authority.
At this point we still need to use that as we're still waiting on [proper certificates](https://github.com/frequenz-io/infrastructure/issues/73)

So this is blocked by either:
* BaseClient supporting custom root certificates.
* Infra giving us an officially recognized SSL certificate.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,13 +2,16 @@
 
 ## Summary
 
-This release adds support for pagination in the dispatch list request.
+This release includes a new feature for pagination support in the dispatch list request as well as usage of the base-client for setting up the channel and client configuration.
 
 ## Upgrading
 
 - The `Client.list()` function now yields a `list[Dispatch]` representing one page of dispatches
+- `Client.__init__` no longer accepts a `grpc_channel` argument, instead a `server_url` argument is required.
 
 ## New Features
 
 - Pagination support in the dispatch list request.
-
+- `Client.__init__`:
+ - Has a new parameter `connect` which is a boolean that determines if the client should connect to the server on initialization.
+ - Automatically sets up the channel for encrypted TLS communication.
--- a/src/frequenz/client/dispatch/_client.py
+++ b/src/frequenz/client/dispatch/_client.py
@@ -3,9 +3,9 @@
 
 """Dispatch API client for Python."""
 from datetime import datetime, timedelta
+from importlib.resources import files
+from pathlib import Path
 from typing import Any, AsyncIterator, Awaitable, Iterator, cast
-
-import grpc
 
 # pylint: disable=no-name-in-module
 from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams
@@ -27,6 +27,8 @@ from frequenz.api.dispatch.v1.dispatch_pb2 import (
     UpdateMicrogridDispatchResponse,
 )
 
+from frequenz.client.base.channel import ChannelOptions, SslOptions
+from frequenz.client.base.client import BaseApiClient
 from frequenz.client.base.conversion import to_timestamp
 
 from ._internal_types import DispatchCreateRequest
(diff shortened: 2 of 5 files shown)


llucax:
> BaseClient supporting custom root certificates

Can you create an issue (or better a PR :innocent: ) with what's needed? An example of how you are creating the grpc channel would help.

Marenz:
You can see the according code here: https://github.com/frequenz-floss/frequenz-client-dispatch-python/blob/v0.x.x/src/frequenz/client/dispatch/__main__.py#L57
We add a repo-local certificate here as root so that our client accepts the self-signed certificate of the service.

llucax:
So looking at that now, and you seem to have a function that accepts more than just the root cert, so for now I will only implement passing a root cert to the base client. Let me know if I am misreading this and you also need to pass a private key and a cert chain.

llucax:
Also, any reason for storing the root cert in the repo and using `files(...)`? Shouldn't the root cert be provided by whoever actually instantiates the client and not the client package itself?

llucax:
In any case, the base client will take the root cert as a path to a file and let clients decide how to do the config, but it seems weird to ship a root cert and force any client users to use that root cert.

Marenz:
> Shouldn't the root cert be provided by whoever actually instantiates the client and not the client package itself?

no, because they wouldn't have the cert ;) 

We generated the cert, we signed the server certs with it and we also ship it in the client because that's the only way they can validate the server cert.

llucax:
> We generated the cert, we signed the server certs with it and we also ship it in the client because that's the only way they can validate the server cert.

You can publish the certs somewhere, like you do with you gpg public key, no? Anyway:

- https://github.com/frequenz-floss/frequenz-client-base-python/pull/73

llucax:
I might need to understand the cert part better, as if the cert really needs to be shipped by the client then passing it via URL doesn't make a lot of sense...

Marenz:
> You can publish the certs somewhere, like you do with you gpg public key, no? Anyway:

I mean sure, but why give our client users so much work if we can just ship it straight with the client?
It's the only place were its used and useful. 
Remember also that this is a temporary solution until infrastructure gives me a proper certificate signed by a real CA... https://github.com/frequenz-io/infrastructure/issues/73

llucax:
Yeah, ok, given let's encrypt I hope we can get normal certificates without needing a root cert. Not sure if we should event merge the PR, maybe for "advanced use" the client should just accept a pre created channel instead of a URL.

src/frequenz/client/dispatch/_client.py
```diff
@@ -37,23 +40,43 @@
...
+            svc_addr,
+            dispatch_pb2_grpc.MicrogridDispatchServiceStub,
+            connect=connect,
+            channel_defaults=ChannelOptions(
+                port=DEFAULT_DISPATCH_PORT,
+                ssl=SslOptions(
+                    enabled=True,
+                    root_certificates=Path(
+                        str(
+                            files("frequenz.client.dispatch").joinpath("certs/root.crt")
```
llucax, line 73: Doesn't this already return a `Path`?

Marenz: Unfortunately, `joinpath` seems to return a `Traversable`

llucax:
```py
>>> import pathlib
>>> p = pathlib.Path("/tmp").joinpath("lala/papap")
>>> p
PosixPath('/tmp/lala/papap')
>>> isinstance(p, pathlib.Path)
True
```
:thinking:

llucax: And:
```py
>>> p2 = importlib.resources.files("frequenz.sdk")
>>> p2
PosixPath('/home/luca/devel/sdk/src/frequenz/sdk')
>>> isinstance(p2, pathlib.Path)
True
```

So, a `Path` is probably a `Traversable`, but it is still a `Path` :D

llucax: Oh, but if it is typed as `Traversable` you would have to cast it I guess. Nevermind, all good, nothing to see here.

src/frequenz/client/dispatch/_client.py
```diff
@@ -37,23 +40,43 @@
...
 
-class Client:
+class Client(BaseApiClient[dispatch_pb2_grpc.MicrogridDispatchServiceStub]):
     """Dispatch API client."""
 
     def __init__(
-        self, *, grpc_channel: grpc.aio.Channel, svc_addr: str, key: str
+        self,
+        *,
+        svc_addr: str,
```
llucax: Sorry I missed this before, but could you use `service_url` instead, like the base client, so we have a consistent interface across all the different API clients?
```suggestion
        server_url: str,
```

llucax (commented):
Sorry, last change request (don't forget to update release notes too). :innocent:

RELEASE_NOTES.md
```diff
@@ -2,13 +2,16 @@
 
 ## Summary
 
-This release adds support for pagination in the dispatch list request.
+This release includes a new feature for pagination support in the dispatch list request as well as usage of the base-client for setting up the channel and client configuration.
 
 ## Upgrading
 
 - The `Client.list()` function now yields a `list[Dispatch]` representing one page of dispatches
+- `Client.__init__` no longer accepts a `grpc_channel` argument, instead a `svc_addr` argument is required.
```
llucax:
```suggestion
- `Client.__init__` no longer accepts a `grpc_channel` argument, instead a `server_url` argument is required.
```
image

Marenz: that's what you meant :laughing: 
I updated the release notes while rebasing ;)-

llucax (approved):
:tada:

PR #74, opened 2024-08-14
Implement Pagination for the python client

Marenz:
- **Implement pagination for list requests**
- **Refactor client.list to be more readable**

### DIFF
--- a/src/frequenz/client/dispatch/_client.py
+++ b/src/frequenz/client/dispatch/_client.py
@@ -6,9 +6,10 @@ from datetime import datetime, timedelta
 from typing import Any, AsyncIterator, Awaitable, Iterator, cast
 
 import grpc
-from frequenz.api.dispatch.v1 import dispatch_pb2_grpc
 
 # pylint: disable=no-name-in-module
+from frequenz.api.common.v1.pagination.pagination_params_pb2 import PaginationParams
+from frequenz.api.dispatch.v1 import dispatch_pb2_grpc
 from frequenz.api.dispatch.v1.dispatch_pb2 import (
     CreateMicrogridDispatchResponse,
     DeleteMicrogridDispatchRequest,
@@ -67,8 +68,20 @@ class Client:
         end_to: datetime | None = None,
         active: bool | None = None,
         dry_run: bool | None = None,
+        page_size: int | None = None,
+        page_token: str | None = None,
     ) -> AsyncIterator[Dispatch]:
         """List dispatches.
+
+        This function handles pagination internally and returns an async iterator
+        over the dispatches. Pagination parameters like `page_size` and `page_token`
+        can be used, but they are mutually exclusive.
+
+        The `page_token` argument can be used to continue iterating from a
+        specific token, however there is currently no way to get the internal
+        next page token, so this parameter is mainly for internal use.
+
+        Create an issue if you need this feature.
 
         Example usage:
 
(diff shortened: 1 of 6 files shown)


src/frequenz/client/dispatch/_client.py
```diff
@@ -91,10 +104,18 @@ async def list(
...
+            page_token: optional, token for the next page, will continue iterating from this token.
 
         Returns:
             An async iterator of dispatches.
+
+        Raises:
+            ValueError: If both page_size and page_token are provided.
         """
+        if page_size is not None and page_token is not None:
+            raise ValueError("page_size and page_token are mutually exclusive")
```
llucax: You can also use an overload to detect misuses with `mypy` (I also strongly recommend making arguments keyword only with so many `datetime`/`int`/`bool`/`str`).

I think it should be something like this:
```py
    @overload
    async def list(
        self,
        microgrid_id: int,
        component_selectors: Iterator[ComponentSelector] = iter(()),
        *,
        start_from: datetime | None = None,
        start_to: datetime | None = None,
        end_from: datetime | None = None,
        end_to: datetime | None = None,
        active: bool | None = None,
        dry_run: bool | None = None,
        page_size: int | None = None,
        page_token: Literal[None],
    ) -> AsyncIterator[Dispatch]:

    async def list(
        self,
        microgrid_id: int,
        component_selectors: Iterator[ComponentSelector] = iter(()),
        *,
        start_from: datetime | None = None,
        start_to: datetime | None = None,
        end_from: datetime | None = None,
        end_to: datetime | None = None,
        active: bool | None = None,
        dry_run: bool | None = None,
        page_size: Literal[None],
        page_token: str | None = None,
    ) -> AsyncIterator[Dispatch]:

    async def list(
        self,
        microgrid_id: int,
        component_selectors: Iterator[ComponentSelector] = iter(()),
        *,
        start_from: datetime | None = None,
        start_to: datetime | None = None,
        end_from: datetime | None = None,
        end_to: datetime | None = None,
        active: bool | None = None,
        dry_run: bool | None = None,
        page_size: int | None = None,
        page_token: str | None = None,
    ) -> AsyncIterator[Dispatch]:
```

Marenz:
Updated

src/frequenz/client/dispatch/_client.py
```diff
@@ -91,45 +95,61 @@ async def list(
...
-        )
-        for dispatch in response.dispatches:
-            yield Dispatch.from_protobuf(dispatch)
+        while True:
+            response = await cast(
+                Awaitable[ListMicrogridDispatchesResponse],
+                self._stub.ListMicrogridDispatches(request, metadata=self._metadata),
+            )
+
+            yield [Dispatch.from_protobuf(dispatch) for dispatch in response.dispatches]
```
llucax: You can potentially still do the `Dispatch` object construction lazily if you return a `Page` object. Or simply a generator:

```py
    async def list(...) -> AsyncIterator[Iterator[Dispatch]]:
        ...
        yield (Dispatch.from_protobuf(dispatch) for dispatch in response.dispatches)
```

With this if you do `first_dispatch = next(client.list(...))` only one `Dispatch` object will be ever created. Map will also work:

```py
yield map(Dispatch.from_protobuf, response.dispatches)
```

tests/test_dispatch_client.py
```diff
@@ -81,24 +81,28 @@ async def test_list_dispatches() -> None:
...
     for i in range(100):
         sample = sampler.generate_dispatch()
         await client.create(**to_create_params(1, sample))
 
         dispatch = None
-        async for _dispatch in client.list(microgrid_id=1):
-            dispatch = _dispatch
+        async for page in client.list(microgrid_id=1):
+            for _dispatch in page:
+                dispatch = _dispatch
```
llucax: Unrelated, but you should be able to do this too:

```suggestion
            for dispatch in page:
                pass
```

Variables in `for` in Python outlive their scope.

llucax (dismissed):
Just a couple more optional comments.

src/frequenz/client/dispatch/_client.py
llucax: I feel a bit more strong about this one, I think it makes more sense to return an iterator than a `list` here.

tests/test_dispatch_client.py
Marenz: strictly speaking, `for` variables don't introduce a new scope :P

Marenz:
updated

llucax (approved):

PR #75, opened 2024-08-20
Simplify test code using fixtures

Marenz:
- **Simplify test with fixtures and new client features**


src/frequenz/client/dispatch/test/fixtures.py
```diff
@@ -0,0 +1,28 @@
...
+@fixture
+def sample() -> Dispatch:
+    """Return a dispatch sample."""
+    return DispatchGenerator().generate_dispatch()
+
+
+@fixture
+def generator() -> DispatchGenerator:
+    """Return a dispatch generator."""
+    return DispatchGenerator()
```
llucax: No need to create 2 generators, no?

```suggestion
@fixture
def generator() -> DispatchGenerator:
    """Return a dispatch generator."""
    return DispatchGenerator()


@fixture
def sample(generator: DispatchGenerator) -> Dispatch:
    """Return a dispatch sample."""
    return generator.generate_dispatch()
```

This way if you use both fixtures the generator will be reused.

llucax (approved):

PR #78, opened 2024-08-27
Replace *_HOST and *_PORT env variables with *_URL variable

Marenz:
To be consistent with the other APIs

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,7 @@ This release includes a new feature for pagination support in the dispatch list 
 
 - The `Client.list()` function now yields a `list[Dispatch]` representing one page of dispatches
 - `Client.__init__` no longer accepts a `grpc_channel` argument, instead a `server_url` argument is required.
+- For the dispatch-cli client, `DISPATCH_API_PORT` and `DISPATCH_API_HOST` environment variables have been replaced with `DISPATCH_API_URL` which should be a full URL including the protocol (e.g. `grpc://fz-0004.frequenz.io:50051`)
 
 ## New Features
 
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -32,44 +32,16 @@ from ._cli_types import (
 )
 from ._client import Client
 
-DEFAULT_DISPATCH_API_HOST = "fz-0004.frequenz.io"
-DEFAULT_DISPATCH_API_PORT = 50051
-
-
-def get_client(*, host: str, port: int, key: str, insecure: bool) -> Client:
-    """Get a new client instance.
-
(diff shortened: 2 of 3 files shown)


src/frequenz/client/dispatch/__main__.py
```diff
@@ -32,24 +32,22 @@
 )
 from ._client import Client
 
-DEFAULT_DISPATCH_API_HOST = "fz-0004.frequenz.io"
-DEFAULT_DISPATCH_API_PORT = 50051
+DEFAULT_DISPATCH_API_URL = "grpc://fz-0004.frequenz.io:50051?ssl=true"
```
llucax: Not sure if this is based on code that was upgraded to client-base 0.6.0, but if that's the case now `ssl=true` is the default.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -32,24 +32,22 @@
...
+        url: The URL for the Dispatch API.
         key: The API key for authentication.
         insecure: Allow use of insecure connection.
 
     Returns:
         Client: A new client instance.
     """
     return Client(
-        server_url=f"grpc://{host}:{port}?ssl={not insecure}",
+        server_url=f"grpc://{url}?ssl={not insecure}",
```
llucax: Is this right? Doesn't `url` also include the `grpc://` (and potentially a `?` with some other options)?

Why not just removing the `insecure: bool` option and passing that as part of the URL?

RELEASE_NOTES.md
```diff
@@ -8,6 +8,7 @@ This release includes a new feature for pagination support in the dispatch list
 
 - The `Client.list()` function now yields a `list[Dispatch]` representing one page of dispatches
 - `Client.__init__` no longer accepts a `grpc_channel` argument, instead a `server_url` argument is required.
+- `DISPATCH_API_PORT` and `DISPATCH_API_HOST` environment variables have been replaced with `DISPATCH_API_URL` which should be a full URL including the protocol (e.g. `grpc://fz-0004.frequenz.io:50051?ssl=true`)
```
llucax: If this depends on the version enabling SSL by default, I would remove the `ssl=true` from the release notes too, otherwise people will just probably copy&paste and keep the unnecessarily long URL with `ssl=true`.
```suggestion
- `DISPATCH_API_PORT` and `DISPATCH_API_HOST` environment variables have been replaced with `DISPATCH_API_URL` which should be a full URL including the protocol (e.g. `grpc://fz-0004.frequenz.io:50051`)
```

llucax (dismissed):
One last optional comment, but LGTM.

llucax:
(removed auto-merge in case you want to address the comment before merging)

llucax (approved):

PR #81, opened 2024-08-29
Fix bug where recurrence is not properly unset when not specified

Marenz:


llucax:
Try to rebase this to see if the labeler issue is resolved?

src/frequenz/client/dispatch/_client.py
```diff
@@ -221,7 +221,7 @@ async def create(
             active=active,
             dry_run=dry_run,
             payload=payload or {},
-            recurrence=recurrence or RecurrenceRule(),
+            recurrence=recurrence,
```
llucax, line 224: Is this really necessary? If so could you explain why?

Marenz: The server (currently) expects a valid recurrence object, not a default-constructed one

llucax (approved):

PR #84, opened 2024-09-03
Implement streaming

Marenz:


src/frequenz/client/dispatch/_client.py
```diff
@@ -170,6 +180,48 @@ def to_interval(
...
+                stream_method=lambda: cast(
+                    AsyncIterator[StreamMicrogridDispatchesResponse],
+                    self.stub.StreamMicrogridDispatches(
+                        request, metadata=self._metadata
+                    ),
+                ),
+                transform=DispatchEvent.from_protobuf,
+            )
+
+        return self.streams[microgrid_id]
```
llucax: Very silly optimization to do only one lookup :stuck_out_tongue: 

```suggestion
        broadcaster = self.streams.get(microgrid_id)
        if broadcaster is None:
            request = StreamMicrogridDispatchesRequest(microgrid_id=microgrid_id)
            broadcaster = GrpcStreamBroadcaster(
                stream_name="StreamMicrogridDispatches",
                stream_method=lambda: cast(
                    AsyncIterator[StreamMicrogridDispatchesResponse],
                    self.stub.StreamMicrogridDispatches(
                        request, metadata=self._metadata
                    ),
                ),
                transform=DispatchEvent.from_protobuf,
            )
            self.streams[microgrid_id] = broadcaster

        return broadcaster
```

llucax (dismissed):
Out of curiosity, there is no way to *filter* streams like you can do with list, right?

Super minor suggestion, LGTM!

llucax:
BTW, nice to have dispatch events streaming!

Marenz:
> Out of curiosity, there is no way to filter streams like you can do with list, right?

So far the API provides no fields for that. It wouldn't be hard to add, but I doubt its worth the work tbh.

llucax:
> So far the API provides no fields for that. It wouldn't be hard to add, but I doubt its worth the work tbh.

Do you feel the same about list? For me it feels weird that they are asymmetrical, but maybe I'm missing some fundamental reason why it makes sense for one and not the other.

llucax (approved):

Marenz:
Lists are used in very different contexts, e.g. for the UI to display the dispatches, not to react to them (usually at least) and do something with them.

The ones that use `stream` usually are interested in ALL changes (e.g. the dispatch actor on an edge) for their microgrid_id (which is the only and non-optional filter we have)

PR #85, opened 2024-09-09
Add support for duration=None

Marenz:


src/frequenz/client/dispatch/types.py
```diff
@@ -332,23 +338,26 @@ def to_protobuf(self) -> PBDispatch:
...
+        return PBDispatch(
+            metadata=DispatchMetadata(
+                dispatch_id=self.id,
+                create_time=to_timestamp(self.create_time),
+                modification_time=to_timestamp(self.update_time),
+            ),
+            data=DispatchData(
+                type=self.type,
+                start_time=to_timestamp(self.start_time),
+                duration=int(self.duration.total_seconds()) if self.duration else None,
```
daniel-zullo-frequenz: `round()` was used in `internal_types` to set the duration. You might want to use round here as well for consistency?

tests/test_dispatch_cli.py
```diff
@@ -337,6 +337,7 @@ async def test_create_command(  # pylint: disable=too-many-arguments,too-many-lo
             ),
         )
 
+    print(result.output)
```
daniel-zullo-frequenz: do you want to keep this print()?

tests/test_dispatch_cli.py
```diff
@@ -172,9 +172,9 @@ async def test_list_command(  # pylint: disable=too-many-arguments
                 "create",
                 "829",
                 "test",
+                "BATTERY",
                 "in 1 hour",
                 "1h",
-                "BATTERY",
```
daniel-zullo-frequenz, line 177: Is this change due to the positional arguments change the order as now there might not be a duration? I just want to be sure I got this correctly

daniel-zullo-frequenz (commented):
I have a few questions to check for. LGTM otherwise

tests/test_dispatch_cli.py
Marenz: Yep, this change is due to the fact that it can now be skipped, so I moved it to the end (as those parameters are position based, not named based)

Marenz:
updated

daniel-zullo-frequenz (dismissed):

src/frequenz/client/dispatch/_internal_types.py
```diff
@@ -25,6 +25,8 @@
     component_selector_to_protobuf,
 )
 
+# pylint: enable=no-name-in-module
+
```
llucax, line 29: This doesn't look right here, but you could also just disable it globally in the whole project as `mypy` will check for this anyway.

src/frequenz/client/dispatch/_internal_types.py
```diff
@@ -42,8 +44,11 @@ class DispatchCreateRequest:
     start_time: datetime
     """The start time of the dispatch in UTC."""
 
-    duration: timedelta
-    """The duration of the dispatch, represented as a timedelta."""
+    duration: timedelta | None
+    """The duration of the dispatch, represented as a timedelta.
+
+    If None, the dispatch is considered to be "infinite" or "no-duration".
```
llucax: For me "no-duration" is not super clear, I think I would say "instantaneous" (or maybe both), or use an example as you did in the docs above.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -239,9 +239,9 @@ def validate_reccurance(ctx: click.Context, param: click.Parameter, value: Any)
     required=True,
     type=str,
 )
-@click.argument("start-time", required=True, type=FuzzyDateTime())
-@click.argument("duration", required=True, type=FuzzyTimeDelta())
 @click.argument("selector", required=True, type=SelectorParamType())
+@click.argument("start-time", required=True, type=FuzzyDateTime())
+@click.argument("duration", required=False, type=FuzzyTimeDelta())
```
llucax, line 244: In the second commit there are quite a few reordering changes that don't seem related, here and also in many imports or string lists.

llucax (commented):
LGTM besides the small optional comments. The only thing that I would probably add is a test for receiving a dispatch with no duration, not sure if it is tested indirectly in the CLI test.

src/frequenz/client/dispatch/_internal_types.py
Marenz: I think this is mostly at this place due to `black`

src/frequenz/client/dispatch/__main__.py
Marenz: Yeah, I blame black.

Marenz: Ah no, the reordering here is very much related. `duration` became an optional positional parameter, so it had to move to the end

src/frequenz/client/dispatch/_internal_types.py
llucax: Yeah, it is weird, it has to be just before the next line.

src/frequenz/client/dispatch/__main__.py
llucax: Really, if black is reordering that it should be a nasty bug, decorators are not commutable, if you change the order the results could be completely different!

daniel-zullo-frequenz (approved):

PR #86, opened 2024-09-12
Fix duration in cli.create

Marenz:
- **Fix duration in cli client.create**
- **Cli: Echo URL used for requests on stderr**

### DIFF
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -58,6 +58,8 @@ async def cli(ctx: click.Context, url: str, key: str) -> None:
     if ctx.obj is None:
         ctx.obj = {}
 
+    click.echo(f"Using API URL: {url}", err=True)
+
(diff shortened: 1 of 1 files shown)


Marenz:
no release notes as we already mention the duration fixes and there was no new release

daniel-zullo-frequenz (approved):

src/frequenz/client/dispatch/__main__.py
```diff
@@ -58,6 +58,8 @@ async def cli(ctx: click.Context, url: str, key: str) -> None:
     if ctx.obj is None:
         ctx.obj = {}
 
+    click.echo(f"Using API URL: {url}", err=True)
```
llucax, line 61: Just a minor suggestion for the future: it might make sense to have some sort of `--verbose` or `--debug` flag to enable this, it seems a bit too verbose for the default case. You can even use the logging system to implement this.

PR #94, opened 2024-09-25
Client.stream: Raise exception on connection loss

Marenz:


src/frequenz/client/dispatch/_client.py
```diff
@@ -208,6 +209,11 @@ def _get_stream(
     ) -> GrpcStreamBroadcaster[StreamMicrogridDispatchesResponse, DispatchEvent]:
         """Get an instance to the streaming helper."""
         broadcaster = self.streams.get(microgrid_id)
+        # pylint: disable=protected-access
+        if broadcaster is not None and broadcaster._channel.is_closed:
```
llucax, line 213: This is not nice. We should add a `broadcaster.is_running()` or something. I think there is no one day that goes by without me thinking "we need to implement this using [`core.asyncio.Service`](https://frequenz-floss.github.io/frequenz-core-python/latest/reference/frequenz/core/asyncio/#frequenz.core.asyncio.Service)". :grimacing: 

After I finish with the microgrid API client... :pray:

llucax (approved):
I'm fine with merging this as is and update when the `client-base` gets any needed changes.

PR #96, opened 2024-10-02
Include init files in sybil

Marenz:


llucax:
This repo doesn't have any examples in the `__init__.py` files though. Not sure if the issue happened also in this case.

llucax (approved):
LGTM but it would be interesting to try this in a repo that has examples in the `__init__.py` files.

PR #99, opened 2024-10-30
Run previously failed tests first

Marenz:
- **CI: Run previously failed tests first**
- **CI: Use uv backend for nox**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,8 +57,8 @@ jobs:
 
       - name: Install required Python packages
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install -e .[dev-noxfile]
+          python -m pip install --upgrade pip uv
+          python -m uv pip install -e .[dev-noxfile]
           pip freeze
 
       # Restore pytest cache only for pytest_min session to use the cache
@@ -75,7 +75,7 @@ jobs:
       - name: Create nox venv
         env:
           NOX_SESSION: ${{ matrix.nox-session }}
-        run: nox --install-only -e "$NOX_SESSION"
+        run: nox -db uv --install-only -e "$NOX_SESSION"
 
       - name: Print pip freeze for nox venv (debug)
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -74,12 +85,26 @@ jobs:
...
+          if [ "$NOX_SESSION" = "pytest_min" ]; then
+            nox -db uv -R -e "$NOX_SESSION" -- --ff -o cache_dir=.pytest_cache
+          else
+            nox -db uv -R -e "$NOX_SESSION"
+          fi
         timeout-minutes: 10
 
+      # Save pytest cache only for pytest_min session
+      - name: Save pytest cache
+        if: always() && matrix.nox-session == 'pytest_min'
```
daniel-zullo-frequenz, line 102: I think `matrix.nox-session` will be interpreted as a string rather  than being evaluated as a variable without `${{ }}`, thus the condition will be always false
```suggestion
        if: always() && ${{ matrix.nox-session == "pytest_min" }}
```

Marenz: No, it's actually working, I tested this PR extensively :)

image

daniel-zullo-frequenz: ok, I see it works.
I read the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution) too and that confirms it too, though there are exceptions for cases that this won't work.
```
When you use expressions in an if conditional, you can, optionally, omit the ${{ }} expression syntax because GitHub Actions automatically evaluates the if conditional as an expression. However, this exception does not apply everywhere.

You must always use the ${{ }} expression syntax or escape with '', "", or () when the expression starts with !, since ! is reserved notation in YAML format. For example:
```

I've also seen in line 66 it is written `if: ${{ matrix.nox-session == 'pytest_min' }}` and ideally we should be consistent on how these statements are written.
IMO it feels safer to always use `${{ }}`, and I agree it doesn't look good

llucax (commented):
I would prefer if repos CI don't start to diverge, so I would definitely push for moving this to a common place like repo-config, or even better, https://github.com/frequenz-floss/gh-action-nox, and start pushing to use these GH actions more.

As for `uv`, I still wonder if pre-releases won't be an issue to start using it as the default. But maybe we have to suck it up and adapt to whatever `uv` wants, as it is orders of magnitude better than `pip`.

Marenz:
> pushing to use these GH actions more.

What you mean "more" I looked and there isn't a single repo using them :P
Could you make one so we have a nice example of how it's used?

llucax:
Internal repo, but there is an usage example in the action repo itself.

PR #100, opened 2024-10-30
Enhance readme, add links, reformat

Marenz:


README.md
```diff
@@ -1,22 +1,49 @@
...
+```
+
+## 🛠️ Usage
+
+Here's a quick example to get you started:
+
+```python
+from frequenz.client.dispatch import Client
+
+# Initialize the client
```
daniel-zullo-frequenz: I'd suggest to wrap all the code within an async function. I know this will make the example much more verbose but otherwise the user might try to instantiate the client outside the asynchronous event loop.

For instance:
```python
import asyncio
from frequenz.client.dispatch import Client

async def list_dispatches() -> None:
    # Initialize the client
    client = Client(key="your_api_key", server_url="grpc://fz-0004.frequenz.io")

    # List all dispatches for a specific microgrid
    async for page in client.list(microgrid_id=1):
        for dispatch in page:
            print(dispatch)

asyncio.run(list_dispatches())
```

daniel-zullo-frequenz (commented):
I have only a suggestion to check for. LGTM otherwise

daniel-zullo-frequenz (approved):

llucax (commented):
Part of this is automatically generated by the repo config template, it would be good to update the template to put everything that is common to all repos there, to avoid every project diverging into completely different README structure and information.

llucax:
@Marenz :point_up:

PR #105, opened 2024-11-07
Update API and add support for start_immediately

Marenz:
- **Update API dependency**
- **CLI: Change interval default to 1**
- **Add support for start_immediately**

### DIFF
--- a/src/frequenz/client/dispatch/_cli_types.py
+++ b/src/frequenz/client/dispatch/_cli_types.py
@@ -130,10 +130,10 @@ class FuzzyIntRange(click.ParamType):
             self.fail(f"Invalid integer range: {value}", param, ctx)
 
 
-class SelectorParamType(click.ParamType):
-    """Click parameter type for selectors."""
+class TargetComponentParamType(click.ParamType):
+    """Click parameter type for targets."""
 
-    name = "selector"
+    name = "target"
 
     def convert(
         self, value: Any, param: click.Parameter | None, ctx: click.Context | None
@@ -154,7 +154,7 @@ class SelectorParamType(click.ParamType):
         values = value.split(",")
 
         if len(values) == 0:
-            self.fail("Empty selector list", param, ctx)
+            self.fail("Empty target list", param, ctx)
 
         error: Exception | None = None
         # Attempt to parse component ids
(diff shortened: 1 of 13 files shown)


src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -32,12 +32,15 @@ def __init__(self) -> None:
...
     def convert(
         self, value: Any, param: click.Parameter | None, ctx: click.Context | None
-    ) -> datetime:
-        """Convert the value to a datetime object."""
-        if isinstance(value, datetime):
-            return value
+    ) -> datetime | Literal["NOW"] | None:
+        """Convert the value to a datetime object or the string "NOW"."""
+        if isinstance(value, datetime) or value == "NOW":
+            return cast(datetime | Literal["NOW"], value)
```
llucax: Since you are also returning `Literal["NOW"]` below to do a more relaxed comparison to `"NOW"`, wouldn't it be easier and avoid the cast to just do this here?
```suggestion
        if isinstance(value, datetime):
            return value
```

src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -32,12 +32,15 @@ def __init__(self) -> None:
...
 
         try:
+            if value in ("NOW", "now"):
```
llucax:
```suggestion
            if value.upper() == "NOW":
```

llucax (dismissed):
LGTM, just 2 small optional suggestions.

I would use a [sentinel](https://peps.python.org/pep-0661/)[^1] instead of `Literal["NOW"]`, it seems less error prone and less typing, but this was not introduced by this PR, so for a separate PR I guess.

[^1]: I think we should add the reference implementation to [`core`](https://github.com/frequenz-floss/frequenz-core-python/).

llucax:
> 1. I think we should add the reference implementation to [`core`](https://github.com/frequenz-floss/frequenz-core-python/). [↩](#user-content-fnref-1-74beafc834d8b89b1ab3d6dc83f02cd9)

* https://github.com/frequenz-floss/frequenz-core-python/pull/37

src/frequenz/client/dispatch/_cli_types.py
Marenz: This implies the type is a string though

PR #106, opened 2024-11-18
Update readme for release

Marenz:


Marenz:
Deletes an entry from the previous release

PR #109, opened 2024-11-18
Use new BaseClient version

Marenz:
- **Reset release notes after release**
- **Update dependency frequenz-client-base to 0.8.0**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ requires-python = ">= 3.11, < 4"
 dependencies = [
   "typing-extensions >= 4.6.1, < 5",
   "frequenz-api-dispatch == 1.0.0-rc1",
-  "frequenz-client-base >= 0.7.0, < 0.8.0",
+  "frequenz-client-base >= 0.8.0, < 0.9.0",
   "frequenz-client-common >= 0.1.0, < 0.3.0",
   "grpcio >= 1.66.1, < 2",
 ]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ requires-python = ">= 3.11, < 4"
 dependencies = [
   "typing-extensions >= 4.6.1, < 5",
   "frequenz-api-dispatch == 1.0.0-rc1",
-  "frequenz-client-base >= 0.7.0, < 0.8.0",
+  "frequenz-client-base >= 0.8.0, < 0.9.0",
   "frequenz-client-common >= 0.1.0, < 0.3.0",
```
ela-kotulska-frequenz, line 42: Why don't you update RELEASE_NOTES?

llucax (commented):
Same question as Ela, otherwise LGTM. Feel free to merge if you add notes or you think notes are not necessary because you plan to invalidate the previous version or something.

pyproject.toml
Marenz: Hmm I usually update them if it's something the user can see, but nothing changed from their perspective in this case, just an indirect dependency

llucax: Yeah, the only issue with bumping dependencies with breaking changes is users are forced to upgrade the dependency too (the thing that is triggering all these release trains), which will force them to update the code because the changes are breaking. This is why we usually document *disruptive* dependency bumps, because it has a high potential of actually affecting the user at the end. But in development mode (v0.x.x) I think is not that bad if it is missing in the release notes, people are supposed to expect breakage anyway.

Marenz: Fair

PR #110, opened 2024-11-27
Add colors and formatting to dispatch-cli

Marenz:
color-dispatch

### DIFF
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -5,6 +5,7 @@
 
 import asyncio
 import os
+from datetime import datetime, timezone
 from pprint import pformat
 from typing import Any, List
 
@@ -16,13 +17,6 @@ from prompt_toolkit.history import FileHistory
 from prompt_toolkit.patch_stdout import patch_stdout
 from prompt_toolkit.shortcuts import CompleteStyle
 
-from frequenz.client.dispatch.recurrence import (
-    EndCriteria,
-    Frequency,
-    RecurrenceRule,
-    Weekday,
-)
-
 from ._cli_types import (
     FuzzyDateTime,
     FuzzyIntRange,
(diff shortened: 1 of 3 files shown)


src/frequenz/client/dispatch/__main__.py
```diff
@@ -31,10 +25,140 @@
...
+    """Return the name of the month."""
+    return datetime(2000, month, 1).strftime("%B")
+
+
+# pylint: disable=too-many-statements, too-many-locals
+def print_dispatch(dispatch: Dispatch) -> None:
+    """Print the dispatch details in a nicely formatted way with colors."""
+    # Determine the status and color
+    status: str = "running" if dispatch.started else "not running"
+    status_color: str = "green" if dispatch.started else "red"
```
llucax, line 74: Nitpick: What about gray for not started? Red looks like an error to me. Maybe we can sort of mimic `systemctl status` as many users may be familiar with them?

src/frequenz/client/dispatch/__main__.py
```diff
@@ -31,10 +25,140 @@
...
+    ]
+    max_key_length: int = max(len(k) for k in keys)
+
+    # Helper function to format each line
+    def format_line(key: str, value: str, color: str = "cyan") -> str:
+        key_str = click.style(f"{key}:", fg=color)
+        val_color = "white"
+
+        if value in ("None", "False"):
+            val_color = "red"
```
llucax, line 120: Also here maybe gray for `None`?

src/frequenz/client/dispatch/__main__.py
```diff
@@ -52,8 +176,15 @@
     show_envvar=True,
     required=True,
 )
+@click.option(
+    "--raw",
```
llucax, line 180: Possible future improvements:
* Split `--raw` from `--no-color`
* Set color on/off automatically based on `sys.stdin.isatty()`

llucax (approved):
Awesome! It might also be a good idea to maybe set constants with colors, so they are easier to tune if needed, like `COLOR_DISPATCH_STARTED`, `VALUE_NONE_COLOR`, `COLOR_VALUE_TRUE`, etc.

But all comments optional.

src/frequenz/client/dispatch/__main__.py
Marenz: > Set color on/off automatically based on sys.stdin.isatty()

that actually seems to happen automatically, at least there is no color if I pipe it through cat or forward it into a file

src/frequenz/client/dispatch/__main__.py
Marenz: My desire was that it stands out immediately what the status is :)

src/frequenz/client/dispatch/__main__.py
llucax: Ah cool. Extra point for `click` 💯

PR #118, opened 2024-12-03
Fix missing dependency

Marenz:


llucax (commented):
How can it be missing and tests passing anyway? Was it installed transitively by some other dependency?

llucax (approved):
Approving besides the question.

Marenz:
It seems to be added indirectly through the `dev` dependencies

PR #124, opened 2025-01-21
Fix some problems that related to flakey tests

Marenz:
- **Fix last_id in fake service when setting dispatches**
- **Always recreate the stream instance**

### DIFF
--- a/src/frequenz/client/dispatch/_client.py
+++ b/src/frequenz/client/dispatch/_client.py
@@ -55,11 +55,6 @@ DEFAULT_DISPATCH_PORT = 50051
 class Client(BaseApiClient[dispatch_pb2_grpc.MicrogridDispatchServiceStub]):
     """Dispatch API client."""
 
-    streams: dict[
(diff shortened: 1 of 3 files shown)


src/frequenz/client/dispatch/_client.py
```diff
@@ -55,11 +55,6 @@
 class Client(BaseApiClient[dispatch_pb2_grpc.MicrogridDispatchServiceStub]):
     """Dispatch API client."""
 
-    streams: dict[
-        int, GrpcStreamBroadcaster[StreamMicrogridDispatchesResponse, DispatchEvent]
-    ] = {}
-    """A dictionary of streamers, keyed by microgrid_id."""
-
```
llucax, line 62: This is the bug (fix). This is a **class attribute**, it is global, shared by all `Client` instances and lives through the whole process life. This means it will not be cleaned up between tests, so streamers created by different tests will use different loops.

You just need to move this line to the `__init__()` (while you are at it, probably make it private too) as `self._streams: ... = {}`.

llucax (approved):

PR #126, opened 2025-01-29
Fix that duration=0 was sent & received as None

Marenz:


src/frequenz/client/dispatch/__main__.py
```diff
@@ -407,7 +407,7 @@ async def create(
     kwargs = {k: v for k, v in kwargs.items() if v is not None}
 
     # Required for client.create
-    if not kwargs.get("duration"):
+    if "duration" not in kwargs:
         kwargs["duration"] = None
```
llucax: If you are a performance freak you can avoid the double lookup by using `kwargs.setdefault("duration", None)` (you can even omit `None`, as it is the default `default`, but maybe that's just too obscure :P).

llucax (dismissed):

shsms (approved):

src/frequenz/client/dispatch/__main__.py
shsms: > If you are a performance freak 

Rabbit vs Tortoise vs Python

llucax (approved):

PR #141, opened 2025-04-02
Fix dispatch-cli stream trying to print event as dispatch

Marenz:


src/frequenz/client/dispatch/__main__.py
```diff
@@ -253,7 +253,9 @@ async def stream(ctx: click.Context, microgrid_id: int) -> None:
         if ctx.obj["raw"]:
             click.echo(pformat(message, compact=True))
         else:
-            print_dispatch(message)
+            msg = cast(DispatchEvent, message)
+            print(f"Dispatch {msg.event.name}")
+            print_dispatch(message.dispatch)
```
daniel-zullo-frequenz, line 261: Maybe you can use the new DispatchEvent variable `msg`, like `msg.dispatch` instead?
I also wonder if `cast` is really needed given that it seems it's already using `message` as a DispatchEvent here, or is it another type?

daniel-zullo-frequenz (commented):
Just a minor comment/question, and LGTM

src/frequenz/client/dispatch/__main__.py
Marenz: It's not really needed, it was more for proper type support and editor type hint support.
I did intend to use it below though :D

Marenz:
Updated in a more elegant way I think

daniel-zullo-frequenz (approved):

Marenz:
hu, why did the test work now :confused:  I thought my previous PR broke it.. oh well

PR #151, opened 2025-04-28
Fix stream reconnecting

Marenz:


llucax (commented):
I didn't have a look at the code in detail yet, but...

I don't understand if the fix is because of the addition of the retry strategy option, because of the returning of the streamer or both. Also I don't understand how this solves the reconnection problem. It would be nice to clarify in the PR.

Also, unless you really need something from the broadcaster as a user, I would return a `ReceiverFetcher` instead, so we can keep the interface a bit more abstract and not leak the internals, otherwise the streamer class also becomes part of the interface, which needs a 1.0 from client base too.

About the interface change, this is still at 0.x so I don't see why we can't make breaking changes...

Marenz:
Well, the f*o*x is because of the commit starting with "Fix:" :) where as the other commit is the breaking change..

Marenz:
I am not sure, a `ReceiverFetcher` only lets me get a receiver, I can't check if the channel is closed, manually close the connection or any of the other things the Grpcbroadcaster offers...

Marenz:
I removed the interface change now as it seems to require discussion and we need this fix urgently.

llucax:
> I am not sure, a `ReceiverFetcher` only lets me get a receiver, I can't check if the channel is closed, manually close the connection or any of the other things the Grpcbroadcaster offers...

Well, that was my question. The thing is `GrcpBroadcaster` was thought as an internal thing, so most "services" provided by it were done for the people implementing clients, not the end users of clients. I don't want to expose too much stuff to the end user because then we need to commit to the interface.

I guess you mean manually close the stream, not the connection, right? Also looking at the streamer class, it only has a `stop()`, it doesn't provide any way to know if the stream is still open or not. I'm not sure we want to let users close streams, they might not be the only user of the stream, we have to do that with care. If they want to "unsubscribe" they should just `close()` the receiver.

The more I look at it, the more I think there is no point in exposing the streamer. If we need to allow the user to check if a stream is alive, then we need to first implement that somehow, just exposing the internal streamer is not helping.

But yeah, returning a `ReceiverFetcher` doesn't make sense either, I think we should keep it as it is (return `Receiver`).

llucax (approved):

PR #163, opened 2025-05-12
Temp. add official root certs to work with new domain

Marenz:
We now are getting proper domains & certificates, to support
this transition smoothly, we add the "normal" systems root cert
explicitly to our custom root cert that we used up until now.

This is required as using the root-cert option prevents the client from
looking at the normal systems root certificates which we would need
in this case.

Once all clients are updated and the service is updated, we can remove
all extra options for TLS certs.

Signed-off-by: Mathias L. Baumann <mathias.baumann@frequenz.com>


Marenz (commented):
Lgtm

PR #164, opened 2025-05-14
Fix `end_time` not handling `None` values correctly

### LINKED ISSUE #162: Fix flaking tests (probably due to end_time)

Marenz:
fixes #162

### DIFF
--- a/src/frequenz/client/dispatch/types.py
+++ b/src/frequenz/client/dispatch/types.py
@@ -296,7 +296,11 @@ class Dispatch:  # pylint: disable=too-many-instance-attributes
             type=pb_object.data.type,
             create_time=to_datetime(pb_object.metadata.create_time),
             update_time=to_datetime(pb_object.metadata.modification_time),
-            end_time=to_datetime(pb_object.metadata.end_time),
+            end_time=(
+                to_datetime(pb_object.metadata.end_time)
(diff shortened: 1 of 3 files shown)


src/frequenz/client/dispatch/types.py
```diff
@@ -296,7 +296,11 @@ def from_protobuf(cls, pb_object: PBDispatch) -> "Dispatch":
             type=pb_object.data.type,
             create_time=to_datetime(pb_object.metadata.create_time),
             update_time=to_datetime(pb_object.metadata.modification_time),
-            end_time=to_datetime(pb_object.metadata.end_time),
+            end_time=(
+                to_datetime(pb_object.metadata.end_time)
+                if pb_object.metadata.HasField("end_time")
```
llucax, line 301: I think we have this *bug* in **many** places in all API clients. We need to be more careful about optional/non-existing fields... For sure I've seen it in the microgrid API, but I think we are lucky there because most stuff we should check if they exist is always filled by the server, so we are not seeing the effect of the bug there.

Marenz (commented):
LGTM

llucax (commented):
LGTM, but I think the release notes could be improved. It is more useful to describe the user-visible effect of the bug. As a user I have no idea about how this bug manifested.

PR #168, opened 2025-05-19
Allow targeting categories with subtypes

### LINKED ISSUE #139: Improve how to get different types of `Dispatch.target`
### What's needed?

It is difficult to use the current targets, [defined](https://github.com/frequenz-floss/frequenz-client-dispatch-python/blob/9b6c949fed356b02dc515c1d04e8eaf949b2f2d2/src/frequenz/client/dispatch/types.py#L30) as `TargetComponents = list[int] | list[ComponentCategory]`, because the `match` statement (and `isinstance()` can't deal with containers, so we can check that something is a `list` but not that is a `list[int]`. We had [issues](https://github.com/frequenz-floss/frequenz-dispatch-python/issues/68) in the past getting this right and we still do.

### Proposed solution

Use wrapper classes, and then the match syntax should work (I think, untested):

```py
class TargetIds(list[int]): pass
class TargetCategories(list[ComponentCategory]): pass
TargetComponents: TypeAlias = TargetIds | TargetCategories

match components:
    case TargetIds() as component_ids:
        print("components to %s", component_ids)
    case TargetCategories() as categories:
        print("components to %s", categories)
    case unsupported:
        assert_never(unsupported)
```

While we are it, maybe we can even change this to use `frozenset` instead of `list` to make it clear there should be no duplicates.

### Use cases

_No response_

### Alternatives and workarounds

Another option could be to make `TargetComponents` a class and have methods to check which (ids or categories) are really present, but I think it will be more implementation boileplate, and a unsafer usage (as it would be hard to get exhaustion checks in case more ways to specify targets are added in the future).

### Additional context

_No response_

Marenz:
- **Update to latest dispatch-api**
- **Implement extra classes for each target type**

Fixes #139

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,12 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* `TargetComponents` was reworked. It now is a type alias for `TargetIds | TargetCategories`:
+ * `TargetIds` can be used to specify one or more specific target IDs:
+  * `TargetIds(1, 2, 3)` or
+  * `TargetIds(ComponentIds(1), ComponentIds(2), ComponentIds(3))`
+ * `TargetCategories` can be used to specify one or more target categories:
+  * `TargetCategories(ComponentCategory.BATTERY, ComponentCategory.SOLAR)`
 
 ## New Features
 
@@ -14,6 +19,21 @@
 * Every call now has a default timeout of 60 seconds, streams terminate after five minutes. This can be influenced by the two new parameters for`DispatchApiClient.__init__()`:
     * `default_timeout: timedelta` (default: 60 seconds)
     * `stream_timeout: timedelta` (default: 5 minutes)
+* With the new `TargetCategory` class (providing `.category` and `.type`) we can now specify subtypes of the categories:
+ * `ComponentCategory.BATTERY` uses `BatteryType` with possible values: `LI_ION`, `NA_ION`
+ * `ComponentCategory.INVERTER` uses `InverterType` with possible values: `BATTERY`, `SOLAR`, `HYBRID`
+ * `ComponentCategory.EV_CHARGER` uses `EvChargerType`: with possible values `AC`, `DC`, `HYBRID`
+ * A few examples on how to use the new `TargetCategory`:
+    * `TargetCategory(BatteryType.LI_ION)`
+        * `category` is `ComponentCategory.BATTERY`
+        * `type` is `BatteryType.LI_ION`
+    * `TargetCategory(ComponentCategory.BATTERY)`
+        * `category` is `ComponentCategory.BATTERY`
+        * `type` is `None`
+    * `TargetCategories(InverterType.SOLAR)`
+        * `category` is `ComponentCategory.INVERTER`
+        * `type` is `InverterType.SOLAR`
+
 
 ## Bug Fixes
 
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -86,7 +86,8 @@ def print_dispatch(dispatch: Dispatch) -> None:
     # Format the target
     if dispatch.target:
         if len(dispatch.target) == 1:
-            target_str: str = str(dispatch.target[0])
+            (first_element,) = dispatch.target
+            target_str: str = str(first_element)
         else:
             target_str = ", ".join(str(s) for s in dispatch.target)
     else:
--- a/src/frequenz/client/dispatch/_cli_types.py
+++ b/src/frequenz/client/dispatch/_cli_types.py
@@ -5,6 +5,7 @@
 
 import json
 from datetime import datetime, timedelta, timezone
+from itertools import chain
 from typing import Any, Literal, cast
 
 import asyncclick as click
@@ -12,6 +13,14 @@ import parsedatetime  # type: ignore
 from tzlocal import get_localzone
 
 from frequenz.client.common.microgrid.components import ComponentCategory
+from frequenz.client.dispatch.types import (
+    BatteryType,
+    EvChargerType,
+    InverterType,
+    TargetCategories,
+    TargetComponents,
+    TargetIds,
+)
 
 # Disable a false positive from pylint
 # pylint: disable=inconsistent-return-statements
@@ -140,7 +149,7 @@ class TargetComponentParamType(click.ParamType):
 
     def convert(
         self, value: Any, param: click.Parameter | None, ctx: click.Context | None
-    ) -> list[ComponentCategory] | list[int]:
+    ) -> TargetIds | TargetCategories:
         """Convert the input value into a list of ComponentCategory or IDs.
 
         Args:
@@ -149,9 +158,9 @@ class TargetComponentParamType(click.ParamType):
             ctx: The Click context object.
 
         Returns:
-            A list of component ids or component categories.
+            A list of targets, either as component IDs or component categories.
         """
-        if isinstance(value, list):  # Already a list
+        if isinstance(value, TargetComponents):
             return value
 
         values = value.split(",")
@@ -162,20 +171,46 @@ class TargetComponentParamType(click.ParamType):
         error: Exception | None = None
         # Attempt to parse component ids
         try:
-            return [int(id) for id in values]
+            return TargetIds(*[int(id) for id in values])
         except ValueError as e:
             error = e
 
+        def enum_from_str(
+            name: str,
+        ) -> InverterType | BatteryType | EvChargerType | ComponentCategory:
+            """Convert a string to an enum member."""
+            name = name.strip().upper()
+            if name in ComponentCategory.__members__:
+                return ComponentCategory[name]
+            if name in InverterType.__members__:
+                return InverterType[name]
+            if name in BatteryType.__members__:
+                return BatteryType[name]
+            if name in EvChargerType.__members__:
+                return EvChargerType[name]
+            raise KeyError(f"Invalid target specification: {name}")
+
         # Attempt to parse as component categories, trim whitespace
         try:
-            return [ComponentCategory[cat.strip().upper()] for cat in values]
+            return TargetCategories(*[enum_from_str(cat) for cat in values])
         except KeyError as e:
             error = e
+
+        types_str = ", ".join(
+            [f"{type.name}" for type in chain(BatteryType, InverterType, EvChargerType)]
+        )
 
         self.fail(
             f'Invalid component category list or ID list: "{value}".\n'
             f'Error: "{error}"\n\n'
-            "Possible categories: BATTERY, GRID, METER, INVERTER, EV_CHARGER, CHP ",
+            "Valid formats:\n"
+            "- 1,2,3 # A list of component IDs\n"
+            "- METER,INVERTER # A list of component categories\n"
+            "- NA_ION,SOLAR # A list of component category types (category is derived)\n"
+            "Valid categories:\n"
+            f"{', '.join([cat.name for cat in ComponentCategory])}\n"
+            "Valid types:\n"
+            f"{types_str}\n",
             param,
             ctx,
         )
--- a/src/frequenz/client/dispatch/test/generator.py
+++ b/src/frequenz/client/dispatch/test/generator.py
@@ -10,7 +10,16 @@ from frequenz.client.common.microgrid.components import ComponentCategory
 
 from .._internal_types import rounded_start_time
 from ..recurrence import EndCriteria, Frequency, RecurrenceRule, Weekday
-from ..types import Dispatch
+from ..types import (
+    BatteryType,
+    Dispatch,
+    EvChargerType,
+    InverterType,
+    TargetCategories,
+    TargetCategory,
+    TargetComponents,
+    TargetIds,
+)
 
 
 class DispatchGenerator:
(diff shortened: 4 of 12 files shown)


src/frequenz/client/dispatch/__main__.py
```diff
@@ -86,7 +86,8 @@ def print_dispatch(dispatch: Dispatch) -> None:
     # Format the target
     if dispatch.target:
         if len(dispatch.target) == 1:
-            target_str: str = str(dispatch.target[0])
+            (el,) = dispatch.target
+            target_str: str = str(el)
```
Marenz: This change is because `target` is a frozenset now

llucax: OK, but why can't use use `[0]` for something frozen? You should be able, or is it because it is a `set` now?

Also what the hell does `el` stands for? It look it is pretty cryptic. Also if you need to clarify in the PR, maybe it is best to add a comment in the code itself for the future ourselves looking at this code.

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,100 @@
...
+    """Unspecified type of EV charger."""
+
+    EV_CHARGER_TYPE_AC = PBEvChargerType.EV_CHARGER_TYPE_AC
+    """AC EV charger."""
+
+    EV_CHARGER_TYPE_DC = PBEvChargerType.EV_CHARGER_TYPE_DC
+    """DC EV charger."""
+
+    EV_CHARGER_TYPE_HYBRID = PBEvChargerType.EV_CHARGER_TYPE_HYBRID
+    """Hybrid EV charger."""
```
llucax: I have these in the microgrid API too, next step is to move them to `client-common` so we can share it. I will do a PR with this soon after I finish the microgrid API 0.17 PR, probably this week. But just FYI, I think we should move forward adding it where needed locally as it might take some time to agree on something we are all happy about to put in common + release. In any case, I would remove the `EV_CHARGER_TYPE_` prefix here.

src/frequenz/client/dispatch/test/generator.py
```diff
@@ -66,6 +76,30 @@ def generate_recurrence_rule(self) -> RecurrenceRule:
             ],
         )
 
+    def generate_target_cat_and_type(self) -> CategoryAndType:
```
llucax: Nitpick: This is already pretty long, I wouldn't save the 5 extra chars from `egory`.

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,100 @@
...
+
+Can be one of the following:
+
+- A set of target component IDs (TargetIds)
+- A set of target component categories (TargetCategories)
+- A set of target component categories and types (TargetCategoryAndTypes)
+
+This is a frozen set, so it is immutable.
+The target components are used to specify the components that a dispatch
+should target.
```
llucax, line 209: Maybe include a `match` example here?

tests/test_proto.py
```diff
@@ -99,7 +130,7 @@ def test_dispatch() -> None:
             start_time=datetime(2024, 10, 10, tzinfo=timezone.utc),
             end_time=datetime(2024, 10, 20, tzinfo=timezone.utc),
             duration=timedelta(days=10),
-            target=[1, 2, 3],
+            target=TargetIds([1, 2, 3]),
```
llucax: Maybe we can add a `__init__()` here to simply allow: `TargetIds(1, 2, 3)`?

tests/test_proto.py
```diff
@@ -120,7 +151,7 @@ def test_dispatch() -> None:
             start_time=datetime(2024, 11, 10, tzinfo=timezone.utc),
             end_time=datetime(2024, 11, 20, tzinfo=timezone.utc),
             duration=timedelta(seconds=20),
-            target=[ComponentCategory.BATTERY],
+            target=TargetCategories([ComponentCategory.BATTERY]),
```
llucax: Same here: `TargetCategories(ComponentCategory.BATTERY)` and for `TargetCategoryAndTypes`, maybe allow something like `TargetCategoryAndTypes(ComponentCategory.BATTERY, InverterType.SOLAR)` (as with the type, we can infer the category, seems to be pretty redundant to require both).

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,100 @@
...
+
+@dataclass(frozen=True)
+class CategoryAndType:
+    """Represents a category and type of dispatch."""
+
+    category: ComponentCategory
+    """The category of the dispatch."""
+
+    type: BatteryType | EvChargerType | InverterType | None
+    """The type of the dispatch."""
```
llucax: This is probably for the API repo, but it seems we now have overlapping `target` options:

```py
TargetCategories([cat1, cat2]) == TargetCategoryAndTypes([CategoryAndType(cat1), CategoryAndType(cat2)])
```
(figuratively speaking)

As for the high-level user interface, since as soon as we have a `type`, the `category` can be inferred, I would simplify for the end user:

```py
# Could still be CategoryAndType, but it sounds a bit weird type is not always present
@dataclass(frozen=True)
class Group:
    target: ComponentCategory | BatteryType | EvChargerType | InverterType

    @property
    def category(self) -> ComponentCategory:
        match self.target:
            case ComponentCategory():
                return self.target
            case BattertyType():
                return ComponentCategory.BATTERY
            ...

    @property
    def type(self) -> ComponentCategory:
        match self.target:
            case ComponentCategory():
                return None
            case type:
                return type
            ...
```

And then add a ctor to class `TargetCategoryAndTypes` to allow for:
```py
TargetCategoryAndTypes(ComponentCategory.BATTERY, InverterType.SOLAR)
```
to result in:
```py
frozenset([Group(ComponentCategory.BATTERY), Group(InverterType.SOLAR)])
```

I'm not sure if we already touched on the redundancy between [`CategorySet`](https://github.com/frequenz-floss/frequenz-api-dispatch/blob/4274d1a30bcb545a7cabf7f4ac542c1a432da981/proto/frequenz/api/dispatch/v1/dispatch.proto#L340-L344) and [`CategoryTypeSet`](https://github.com/frequenz-floss/frequenz-api-dispatch/blob/4274d1a30bcb545a7cabf7f4ac542c1a432da981/proto/frequenz/api/dispatch/v1/dispatch.proto#L346-L374) in the protocol, if not maybe we should.

src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -5,13 +5,23 @@
 
 import json
 from datetime import datetime, timedelta, timezone
+from sre_constants import IN
```
llucax: I bet this was Tab-oriented-programming + auto-import :laughing: 
```suggestion
```

llucax (commented):
LGTM in general, although not completely convinced about the current approach with `CategoryAndType`, looks a bit too low-level for the "high-level interface".

src/frequenz/client/dispatch/__main__.py
Marenz: You cannot index-access a set, it has no order :)

`el` stands for `element`.. sure I'll make it more clear :smile: 
I was surprised as well, but this seems to be the easiest way to access the only element of a set with size 1

src/frequenz/client/dispatch/test/generator.py
Marenz: You're worse than copilot

src/frequenz/client/dispatch/types.py
Marenz: > I'm not sure if we already touched on the redundancy between [CategorySet](https://github.com/frequenz-floss/frequenz-api-dispatch/blob/4274d1a30bcb545a7cabf7f4ac542c1a432da981/proto/frequenz/api/dispatch/v1/dispatch.proto#L340-L344) and [CategoryTypeSet](https://github.com/frequenz-floss/frequenz-api-dispatch/blob/4274d1a30bcb545a7cabf7f4ac542c1a432da981/proto/frequenz/api/dispatch/v1/dispatch.proto#L346-L374) in the protocol, if not maybe we should.

We don't need to talk about those redundancies because they are deprecated and scheduled to be removed, the whole typeless category and it's field in the `oneof` construct will be gone once all clients are a capable of reading/writing both.

src/frequenz/client/dispatch/_cli_types.py
Marenz: it was more desperation, but this is a draft after all :P

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,100 @@
...
+
+
+class TargetIds(frozenset[int]):
+    """A set of target component IDs.
+
+    This is a frozen set, so it is immutable.
+    """
+
```
ela-kotulska-frequenz, line 131: Can we make it 
`class TargetIds(frozenset[ComponentId]):`  from `frequenz.client.microgrid.ComponentId`. ?

llucax: Let's wait for that, we should use it here only when it is in `client-common`. We can always convert between `ComponentId` and `int`. I don't want to rush things, because whatever goes to `client-common` will affect everyone, so it might take some time.

RELEASE_NOTES.md
```diff
@@ -6,14 +6,32 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* `TargetComponents` was reworked. It now is a type alias for `TargetIds | TargetCategories`:
+ * `TargetIds` can be used to specify one or more specific target IDs:
+  * `TargetIds(1, 2, 3)` or
```
llucax, line 11: Unless we have some other situation like the one @ela-kotulska-frequenz mentioned and we want this only as a transitional step, I wouldn't allow raw `int`s, the whole point of introducing `ComponentId` is that you can't accidentally pass some random `int` (or some ID for other class of object) where a specific ID is required.

RELEASE_NOTES.md
```diff
@@ -6,14 +6,32 @@
...
+  * `TargetIds(ComponentIds(1), ComponentIds(2), ComponentIds(3))`
```
llucax, line 12: Oh, but this seems to be fake news, right? This one is not supported yet?

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,180 @@
...
-# pylint: enable=no-name-in-module
 from frequenz.client.common.microgrid.components import ComponentCategory
 
 from .recurrence import Frequency, RecurrenceRule, Weekday
 
-TargetComponents = list[int] | list[ComponentCategory]
-"""One or more target components specifying which components a dispatch targets.
 
-It can be a list of component IDs or a list of categories.
+class EvChargerType(IntEnum):
```
llucax: Don't use `IntEnum`, this makes the values instances of `int`, allowing crazy stuff like `abs(EvChargerType.UNSPECIFIED)` or `EvChargerType.AC / EvChargerType.DC` :man_singer:

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,180 @@
...
+                return None
+
+
+class TargetIds(frozenset[int]):
+    """A set of target component IDs.
+
+    This is a frozen set, so it is immutable.
+    """
+
+    def __new__(cls, *ids: SupportsInt) -> "TargetIds":
```
llucax: :scream: you can't define `__init__` becuase it is `frozen`? :cold_face:

llucax: `from typing import Self`:

```suggestion
    def __new__(cls, *ids: SupportsInt) -> Self:
```

Probably it will not be subclasses, but if it is, the type should be the type of the sub-class, not always this one.

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,180 @@
...
+)
+
+
+class TargetCategories(frozenset[TargetCategory]):
+    """A set of target component categories and types.
+
+    This is a frozen set, so it is immutable.
+    """
+
+    def __new__(cls, *categories_input: TargetCategoryInputType) -> "TargetCategories":
```
llucax:
```suggestion
    def __new__(cls, *categories_input: TargetCategoryInputType) -> Self:
```

src/frequenz/client/dispatch/types.py
```diff
@@ -21,16 +32,180 @@
...
+        """
+        # Convert all provided ids to integers before creating the frozenset
+        processed_ids = tuple(int(id_val) for id_val in ids)
+        return super().__new__(cls, processed_ids)
+
+
+# Define the union of types that can be passed to TargetCategories constructor
+TargetCategoryInputType = (
+    TargetCategory | ComponentCategory | BatteryType | InverterType | EvChargerType
+)
```
llucax, line 149: Usually it is a good idea to document type aliases too so they appear in the docs.
```suggestion
)
"""The union of all types that can be used as target categories."""
```

llucax (commented):
Pretty cool!

RELEASE_NOTES.md
Marenz: This is specifically supported (if the type can be cast to int)!

RELEASE_NOTES.md
Marenz: I mean, we're not introducing it in this PR though

src/frequenz/client/dispatch/types.py
Marenz: I guess? THe init ctor doesn't take parameters :)

RELEASE_NOTES.md
llucax: Yeah, this is wrong. It should only accept `ComponentId`, we created that type as a safety measure, we don't want to accept any `int` crap as a `CompentId` unless someone explicitly converts that `int` into a `ComponentId`.

RELEASE_NOTES.md
llucax: OK, I see, so we can keep the conversion to `int` until this repo starts using `ComponentId` (when it is added to `client-common`. Then we only accept `ComponentId`.

src/frequenz/client/dispatch/types.py
llucax: Even weirder.

llucax (dismissed):

Marenz:
Merging as it only had decorative / rebase changes

RELEASE_NOTES.md
```diff
@@ -6,14 +6,34 @@
...
+ * `TargetCategories` can be used to specify one or more target categories:
+  * `TargetCategories(ComponentCategory.BATTERY, ComponentCategory.SOLAR)`
```
llucax, line 14: There is no `ComponentCategory.SOLAR` 😅

Marenz: https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/178

PR #169, opened 2025-05-22
Add timeouts to all api calls

Marenz:


src/frequenz/client/dispatch/_client.py
```diff
@@ -49,23 +49,34 @@
...
 
+    # pylint: disable-next=too-many-arguments
     def __init__(
         self,
         *,
         server_url: str,
         key: str,
         connect: bool = True,
+        call_timeout: timedelta | None = None,
+        stream_timeout: timedelta | None = None,
```
llucax: Why not
```suggestion
        call_timeout: timedelta = timedelta(seconds=60)
        stream_timeout: timedelta = timedelta(minutes=5),
```
and remove the `DEFAULT_XXX` (and the defaults from the docs, the defaults will be shown in the docs anyway like this).

Marenz: I thought we want to avoid to init objects in the default parameters as they aren't constants.
Granted, as we only ever read this value it doesn't matter too much here..

llucax: `datetime` and `timedelta` are immutable (like a `frozen` `dataclass`). If that weren't the case, it would still be a problem:

```py
my_td = timedelta(minutes=1)
client(timeout=my_td)
my_td.set_minutes(10)
```

llucax (approved):

PR #179, opened 2025-06-05
Client.stream(): Return a streamer instead of a receiver

Marenz:
This allows us to use all the options available in the streamer.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,19 @@ While the new TargetCategory class supports subtypes, only reading them is curre
   * `TargetIds(ComponentIds(1), ComponentIds(2), ComponentIds(3))`
  * `TargetCategories` can be used to specify one or more target categories:
   * `TargetCategories(ComponentCategory.BATTERY, ComponentCategory.INVERTER)`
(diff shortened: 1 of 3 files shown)


llucax (commented):
I'm not convinced about this, `GrpcStreamBroadcaster` is a "low-level" class designed to be used internally by clients. Also the `stream()` might be misleading as a name, as it is now only getting you a `GrpcStreamBroadcaster`.

Another option would be to just accept the same arguments as `GrpcStreamBroadcaster.new_receiver()` in your `stream()` method.

On the other hand, if we return a streamer, it means less code duplication, and having a common interface to "configuring" the receiver for all streaming methods. If we want to go that route, maybe we can create a `Protocol` to only expose the parts of the `GrpcStreamBroadcaster` we want to be public. We even already have something like this, the `ReceiverFetcher`, but it more general, so I think the `new_receiver()` it takes no arguments (or maybe only the buffer size).

In any case, I think we need to talk about this more broadly, so all clients provide the same interface. If one client returns a `GrpcStreamBroadcaster` and another one returns a plain receiver, then our clients will become inconsistent.

PR #180, opened 2025-06-12
Fix `stream()` not working with base-client 0.11

Marenz:
This PR ensures the `stream()` helper recreates a stream if the prior one isn’t running, and adds a test to exercise calling `stream()` twice.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -10,6 +10,7 @@ from datetime import timedelta
 from functools import partial
 
 import grpc
+import pytest
 from pytest import raises
 
 from frequenz.client.dispatch.test.client import FakeClient, to_create_params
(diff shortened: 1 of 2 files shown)


llucax:
I took the liberty to change your PR description with the one Copilot created, you really need to work on your PR/commit messages, they are really not helping the reviewers at all. Not you can even ask the AI overlords to do it for you, so it shouldn't be a very complicated task :unamused:

tests/test_client.py
```diff
@@ -309,6 +309,9 @@ async def test_dispatch_stream(client: FakeClient, sample: Dispatch) -> None:
 
     stream = client.stream(microgrid_id)
 
+    # Call function again to test behavior if stream already exists
+    _ = client.stream(microgrid_id)
+
```
llucax: Wouldn't it make sense to create a separate test for the double call? What if there is a bug that affects the `stream()` function in a way that if doesn't work when called only once but works when called twice?

Marenz: I split the test now

llucax (approved):

PR #181, opened 2025-06-12
Update to use BaseId' ids

Marenz:


src/frequenz/client/dispatch/types.py
```diff
@@ -37,6 +38,11 @@
 from .recurrence import Frequency, RecurrenceRule, Weekday
 
 
+@final
+class DispatchId(BaseId, str_prefix="DID"):
+    """A unique identifier for a microgrid."""
```
llucax:
```suggestion
    """A unique identifier for a dispatch."""
```

tests/test_client.py
```diff
@@ -327,16 +328,19 @@ async def expect(dispatch: Dispatch, event: Event) -> None:
...
 
     # Add a new dispatch
     dispatches[1] = await client.create(**to_create_params(microgrid_id, dispatches[1]))
+    await asyncio.sleep(0.1)
     # Expect the second dispatch
     await expect(dispatches[1], Event.CREATED)
 
     # Add a new dispatch
     dispatches[2] = await client.create(**to_create_params(microgrid_id, dispatches[2]))
+    await asyncio.sleep(0.1)
```
llucax: Why are these sleeps being added in the commit changing to use the ID classes?

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -73,6 +73,17 @@ def __init__(self) -> None:
...
+    def refresh_last_id_for(self, microgrid_id: MicrogridId) -> None:
+        """Update last id to be the next highest number."""
+        dispatches = self.dispatches.get(microgrid_id, [])
+
+        if len(dispatches) == 0:
+            return
+
+        self._last_id = DispatchId(
+            max(int(self._last_id), max(int(dispatch.id) for dispatch in dispatches))
+        )
```
llucax: Didn't `max` work directly on IDs? Id's should be [sortable (less-than-comparable)](https://github.com/frequenz-floss/frequenz-core-python/blob/bfa3ef3e7e9bdecafd6fbb21918a651d915793ff/src/frequenz/core/id.py#L152-L178).

llucax (commented):
LGTM, but the commit split is weird and I'm not sure if the new sleeps sneaked in accidentally or not.

I would keep all (and only) ID updates in one commit, and refactors in a different commit (the refactor commit is updating to use ID classes too).

tests/test_client.py
Marenz: The test started to fail without it, probably due to some dependency changes

src/frequenz/client/dispatch/test/_service.py
Marenz: I did get an error, though maybe it was because it was still somehow mixed at that point? Will try

llucax: If we need to add another dunder method to `BaseId` (like `__gt__`) to make it work, I would go for it.

pyproject.toml
```diff
@@ -182,6 +183,7 @@ filterwarnings = [
   # We use a raw string (single quote) to avoid the need to escape special
   # chars as this is a regex
   'ignore:Protobuf gencode version .*exactly one major version older.*:UserWarning',
+  "ignore:Prefix '.*' is already registered.*:UserWarning",
```
llucax: Oh, damn, it is not nice that we are *leaking* this warning and now we'll need to add it to all projects. I would consider converting the warning into a log or just removing the check.

src/frequenz/client/dispatch/_client.py
```diff
@@ -211,24 +214,13 @@ def to_interval(
...
-        ```
-        client = DispatchApiClient(
-            key="key",
-            server_url="grpc://dispatch.url.goes.here.example.com"
-        )
-        async for message in client.stream(microgrid_id=1):
-            print(message.event, message.dispatch)
-        ```
-
         Args:
```
llucax: Why was this removed?

tests/test_client.py
llucax: This is weird, the dependencies that were updated don't seem to have much to do with timing :thinking: 

In any case it would be good to split it in another commit or at least mention it in the commit message.

src/frequenz/client/dispatch/_client.py
Marenz: Ah, I was testing whether it happens without an accidentally commit it

PR #186, opened 2025-07-15
Support secrets and signing

Marenz:
This PR adds support for request signing via an optional secret and propagates that change through tests, the client library, and the CLI.

### DIFF
--- a/src/frequenz/client/dispatch/__main__.py
+++ b/src/frequenz/client/dispatch/__main__.py
@@ -172,10 +172,17 @@ def print_dispatch(dispatch: Dispatch) -> None:
 )
 @click.option(
     "--api-key",
-    help="API key for authentication",
+    help="API key for authentication (deprecated, use --auth-key or DISPATCH_API_AUTH_KEY)",
     envvar="DISPATCH_API_KEY",
     show_envvar=True,
-    required=True,
+    required=False,
+)
+@click.option(
+    "--auth-key",
+    help="API auth key for authentication",
+    envvar="DISPATCH_API_AUTH_KEY",
+    show_envvar=True,
+    required=False,
 )
 @click.option(
     "--sign-secret",
@@ -193,15 +200,28 @@ def print_dispatch(dispatch: Dispatch) -> None:
     default=False,
 )
 @click.pass_context
-async def cli(
-    ctx: click.Context, url: str, api_key: str, sign_secret: str | None, raw: bool
+async def cli(  # pylint: disable=too-many-arguments, too-many-positional-arguments
+    ctx: click.Context,
+    url: str,
+    api_key: str | None,
+    auth_key: str | None,
+    sign_secret: str | None,
+    raw: bool,
 ) -> None:
     """Dispatch Service CLI."""
     if ctx.obj is None:
         ctx.obj = {}
 
+    key = auth_key or api_key
+
+    if not key:
+        raise click.BadParameter(
+            "You must provide an API auth key using --auth-key or "
+            "the DISPATCH_API_AUTH_KEY environment variable."
+        )
+
     click.echo(f"Using API URL: {url}", err=True)
-    click.echo(f"Using API Auth Key: {api_key[:4]}{'*' * 8}", err=True)
+    click.echo(f"Using API Auth Key: {key[:4]}{'*' * 8}", err=True)
 
     if sign_secret:
         if len(sign_secret) > 8:
@@ -211,16 +231,28 @@ async def cli(
         else:
             click.echo("Using API Signing Secret (not shown).", err=True)
 
+    if api_key and auth_key is None:
+        click.echo(
+            click.style(
+                "Deprecation Notice: The --api-key option and the DISPATCH_API_KEY environment "
+                "variable are deprecated. "
+                "Please use --auth-key or set the DISPATCH_API_AUTH_KEY environment variable.",
+                fg="red",
+                bold=True,
+            ),
+            err=True,
+        )
+
     ctx.obj["client"] = DispatchApiClient(
         server_url=url,
-        auth_key=api_key,
+        auth_key=key,
         sign_secret=sign_secret,
         connect=True,
     )
 
     ctx.obj["params"] = {
         "url": url,
-        "api_key": api_key,
+        "auth_key": key,
         "sign_secret": sign_secret,
     }
 
@@ -228,7 +260,7 @@ async def cli(
 
     # Check if a subcommand was given
     if ctx.invoked_subcommand is None:
-        await interactive_mode(url, api_key, sign_secret)
+        await interactive_mode(url, key, sign_secret)
 
 
 @cli.command("list")
@@ -555,7 +587,7 @@ async def repl(
 ) -> None:
     """Start an interactive interface."""
     await interactive_mode(
-        obj["params"]["url"], obj["params"]["api_key"], obj["params"]["sign_secret"]
+        obj["params"]["url"], obj["params"]["auth_key"], obj["params"]["sign_secret"]
     )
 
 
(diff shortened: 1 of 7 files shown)


src/frequenz/client/dispatch/__main__.py
```diff
@@ -619,6 +635,8 @@ async def display_help() -> None:
                 url,
                 "--key",
                 key,
+                "--secret",
+                secret if secret else "",
             ] + click.parser.split_arg_string(user_input)
```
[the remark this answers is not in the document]
Marenz: passing empty is fine

mkdocs.yml
```diff
@@ -103,14 +103,15 @@ plugins:
...
-            merge_init_into_class: false
+            merge_init_into_class: true
             separate_signature: true
             show_category_heading: true
             show_root_heading: true
             show_root_members_full_path: true
             show_signature_annotations: true
             show_source: true
             signature_crossrefs: true
+            members_order: source
```
llucax: I would leave thing the same as they are configured in other repos. I think making this inconsistent across repos will make the docs harder to navigate to users.

mkdocs.yml
```diff
@@ -103,14 +103,15 @@ plugins:
             paths: ["src"]
             docstring_section_style: spacy
             inherited_members: true
-            merge_init_into_class: false
+            merge_init_into_class: true
```
llucax: This usually looks bad when using google-style docs, I tried it in the past and decided for a separate `__init__`. We could reconsider, but we should do it for all repos.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -177,6 +177,14 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
     show_envvar=True,
     required=True,
 )
+@click.option(
+    "--secret",
+    help="API signing secret for authentication",
+    envvar="DISPATCH_API_SECRET",
```
llucax:
```suggestion
    "--sign-secret",
    help="API secret for signing",
    envvar="DISPATCH_API_SIGN_SECRET",
```

src/frequenz/client/dispatch/__main__.py
```diff
@@ -185,29 +193,36 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
     default=False,
 )
 @click.pass_context
-async def cli(ctx: click.Context, url: str, key: str, raw: bool) -> None:
+async def cli(
+    ctx: click.Context, url: str, key: str, secret: str | None, raw: bool
```
llucax:
```suggestion
    ctx: click.Context, url: str, auth_key: str, sign_secret: str | None, raw: bool
```

src/frequenz/client/dispatch/__main__.py
```diff
@@ -185,29 +193,36 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
...
+) -> None:
     """Dispatch Service CLI."""
     if ctx.obj is None:
         ctx.obj = {}
 
     click.echo(f"Using API URL: {url}", err=True)
+    click.echo(f"Using API Key: {key}", err=True)
```
llucax:
```suggestion
    click.echo(f"Using API Auth Key: {auth_key}", err=True)
```

src/frequenz/client/dispatch/__main__.py
```diff
@@ -185,29 +193,36 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
...
+    if secret and len(secret) > 8:
+        click.echo(f"Using API Secret: {secret[:4]}{'*' * 8}", err=True)
```
llucax:
```suggestion
        click.echo(f"Using API Signing Secret: {sign_secret[:4]}{'*' * 8}", err=True)
```

llucax: I would apply the shortening/obscuring always, and do it for the auth key too, it is not very nice to print credentials in plain text to the console, too risky if this is put into a log and sent somewhere.

If you want to improve debugging, you could print the source of the secret, like if it came from an env var or config or where, so the user can go check it out if something is not working, but not sure if that's easy to do with the integrated click env var support.

src/frequenz/client/dispatch/__main__.py
```diff
@@ -574,7 +590,7 @@ async def delete(
         raise click.ClickException("Some deletions failed.")
 
 
-async def interactive_mode(url: str, key: str) -> None:
+async def interactive_mode(url: str, key: str, secret: str | None) -> None:
```
llucax:
```suggestion
async def interactive_mode(url: str, auth_key: str, sign_secret: str | None) -> None:
```

Same for the whole file.

llucax (commented):
LGTM in general, but I have a few comments.

mkdocs.yml
Marenz: Tried to change it a bit after being frustrated with our gen. docs

mkdocs.yml
Marenz: (or it will make this one repo easier and every other stays harder to navigate :P )

src/frequenz/client/dispatch/__main__.py
Marenz: You could argue that the key is public and only the secret is what really authenticates ;)
But less info is always better, true, so will change it

src/frequenz/client/dispatch/__main__.py
Marenz: Uh.. you wanna rename the env variable?
what about the API_KEY then? Consistently it should become `DISPATCH_API_AUTH_KEY`

mkdocs.yml
llucax: For me the side bar is an index, and as an index it should be sorted alphabetically. If you want to sort things in a more organized way, it might be better to have an overview at the module/class documentation.

mkdocs.yml
llucax: Just for the records, I'm always up for better docs! I just invested quite a bit of time to find a combination of options that worked well together. That said, this was a long time ago, and mkdocstrings changed a lot since then, so maybe exploring this again is a good idea, but we should do it considering the wider implications, I just a bit concerned about doing it randomly in a single repo, as it might need to rethink how we write the docs in terms of separation between class and `__init__` docs.

If you want to experiment with it (or you already did), I said make a demo in the SDK meeting.

src/frequenz/client/dispatch/__main__.py
llucax: Yes, that would be the ideal IMHO. If it is too disruptive by now, I would accept both for a transitional period, but again, not sure how possible that is using `click`'s built-in env var management.

Marenz: I mean, it's only used by ppl directly and it's very easy to adjust to it. I can make it print a warning or so if that var is set. It's not like an API. Just a TUI.

llucax: Right, yeah, not that disruptive. Sounds good to me!

src/frequenz/client/dispatch/__main__.py
```diff
@@ -185,29 +200,67 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
...
+
+    if sign_secret:
+        if len(sign_secret) > 8:
+            click.echo(
+                f"Using API Signing Secret: {sign_secret[:4]}{'*' * 8}", err=True
+            )
+        else:
+            click.echo("Using API Signing Secret (not shown).", err=True)
+
+    if api_key and auth_key is not None:
```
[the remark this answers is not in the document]
Marenz: Good point.  Updated

src/frequenz/client/dispatch/__main__.py
```diff
@@ -171,11 +171,26 @@ def format_line(key: str, value: str, color: str = "cyan") -> str:
     show_envvar=True,
 )
 @click.option(
-    "--key",
-    help="API key for authentication",
+    "--api-key",
+    help="API key for authentication (deprecated, use DISPATCH_API_AUTH_KEY)",
```
llucax: Ah, nice, maybe include `--api-key` too.
```suggestion
    help="API key for authentication (deprecated, use --auth-key or DISPATCH_API_AUTH_KEY instead)",
```

llucax (dismissed):
LGTM, only one small suggestion.

llucax:
Removed from the queue in case you want to address it.

PR #197, opened 2025-08-13
refactor: Adapt code to v1alpha8 API of frequenz-client-common

Marenz:
The `frequenz-client-common` dependency has been updated, which introduced breaking changes from the new `v1alpha8` API. This commit adapts the codebase to these changes, resolving the resulting `nox` failures across `mypy`, `pytest`, and formatters.

Key changes include:
- Replacing the deprecated `ComponentCategory` with `v1alpha8`'s `ElectricalComponentCategory`.
- Updating enum members to match the new API, such as `ElectricalComponentCategory.GRID` to `GRID_CONNECTION_POINT` and `InverterType.PV` to `InverterType.SOLAR`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,7 +20,7 @@ This release updates the `frequenz-api-dispatch` dependency to `v1.0.0-rc3` and 
 * Support secrets for signing and verifying messages.
   * Use the new env variable `DISPATCH_API_SIGN_SECRET` to set the secret key.
   * Use the new `sign_secret` parameter in the `DispatchClient` constructor to set the secret key.
-* Added `auth_key` parameter to the `dispatch-cli` and thew env variable `DISPATCH_API_AUTH_KEY` to set the authentication key for the Dispatch API.
+* Added `auth_key` parameter to the `dispatch-cli` and the new variable `DISPATCH_API_AUTH_KEY` to set the authentication key for the Dispatch API.
 
 
 ## Bug Fixes
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,14 +2,21 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+This release updates the `frequenz-api-dispatch` dependency to `v1.0.0-rc3` and adapts the client to the latest API changes. It introduces more specific component categories and new event names while maintaining backward compatibility.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* `InverterType.SOLAR` is now deprecated in favor of `InverterType.PV`. The old name can still be used but will be removed in a future release.
+* The following dependencies were updated:
+  * `frequenz-api-dispatch` to `v1.0.0-rc3`
+  * `frequenz-client-common` to `v0.3.6`
+  * `grpcio` to `v1.72.1`
 
 ## New Features
 
+* **`ElectricalComponentCategory`**: The client now uses the more specific `frequenz.client.common.microgrid.electrical_components.ElectricalComponentCategory` for targeting components.
+  * The new property `TargetCategory.category2` will return an `ElectricalComponentCategory`.
+  * The existing `TargetCategory.category` property is preserved for backward compatibility and returns the corresponding `ComponentCategory`.
 * Support secrets for signing and verifying messages.
   * Use the new env variable `DISPATCH_API_SIGN_SECRET` to set the secret key.
   * Use the new `sign_secret` parameter in the `DispatchClient` constructor to set the secret key.
--- a/src/frequenz/client/dispatch/_cli_types.py
+++ b/src/frequenz/client/dispatch/_cli_types.py
@@ -13,6 +13,9 @@ import parsedatetime  # type: ignore
 from tzlocal import get_localzone
 
 from frequenz.client.common.microgrid.components import ComponentCategory
+from frequenz.client.common.microgrid.electrical_components import (
+    ElectricalComponentCategory,
+)
 from frequenz.client.dispatch.types import (
     BatteryType,
     EvChargerType,
@@ -177,9 +180,17 @@ class TargetComponentParamType(click.ParamType):
 
         def enum_from_str(
             name: str,
-        ) -> InverterType | BatteryType | EvChargerType | ComponentCategory:
+        ) -> (
+            InverterType
+            | BatteryType
+            | EvChargerType
+            | ElectricalComponentCategory
+            | ComponentCategory
+        ):
             """Convert a string to an enum member."""
             name = name.strip().upper()
+            if name in ElectricalComponentCategory.__members__:
+                return ElectricalComponentCategory[name]
             if name in ComponentCategory.__members__:
                 return ComponentCategory[name]
             if name in InverterType.__members__:
@@ -208,7 +219,7 @@ class TargetComponentParamType(click.ParamType):
             "- METER,INVERTER # A list of component categories\n"
             "- NA_ION,SOLAR # A list of component category types (category is derived)\n"
             "Valid categories:\n"
-            f"{', '.join([cat.name for cat in ComponentCategory])}\n"
+            f"{', '.join([cat.name for cat in ElectricalComponentCategory])}\n"
             "Valid types:\n"
             f"{types_str}\n",
             param,
--- a/src/frequenz/client/dispatch/types.py
+++ b/src/frequenz/client/dispatch/types.py
@@ -10,13 +10,13 @@ from enum import Enum
 from typing import Any, Self, SupportsInt, TypeAlias, cast, final
 
 # pylint: enable=no-name-in-module
-from frequenz.api.common.v1.microgrid.components.battery_pb2 import (
+from frequenz.api.common.v1alpha8.microgrid.electrical_components.electrical_components_pb2 import (
     BatteryType as PBBatteryType,
 )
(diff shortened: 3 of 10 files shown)


src/frequenz/client/dispatch/_cli_types.py
```diff
@@ -177,11 +179,11 @@ def convert(
 
         def enum_from_str(
             name: str,
-        ) -> InverterType | BatteryType | EvChargerType | ComponentCategory:
+        ) -> InverterType | BatteryType | EvChargerType | ElectricalComponentCategory:
```
llucax: Shouldn't you accept the old `ComponentCategory` too to make it backwards compatible?

src/frequenz/client/dispatch/types.py
```diff
@@ -81,7 +84,7 @@ class InverterType(Enum):
     BATTERY = PBInverterType.INVERTER_TYPE_BATTERY
     """Battery inverter."""
 
-    SOLAR = PBInverterType.INVERTER_TYPE_SOLAR
+    PV = PBInverterType.INVERTER_TYPE_PV
```
llucax, line 106: Also probably we need to keep the old name for backwards compatibility.

src/frequenz/client/dispatch/types.py
```diff
@@ -92,28 +95,28 @@ class InverterType(Enum):
...
-    target: ComponentCategory | BatteryType | EvChargerType | InverterType
+    target: ElectricalComponentCategory | BatteryType | EvChargerType | InverterType
     """The target category of the dispatch.
 
     Implicitly derived from the types.
     """
 
     @property
-    def category(self) -> ComponentCategory:
+    def category(self) -> ElectricalComponentCategory:
```
llucax: OK, this is tricky, if we want backwards compatibility we need a new name, like `category2()`, and a conversion to the old enum.

src/frequenz/client/dispatch/types.py
```diff
@@ -151,7 +154,11 @@ def __new__(cls, *ids: SupportsInt) -> Self:
 
 # Define the union of types that can be passed to TargetCategories constructor
 TargetCategoryInputType = (
-    TargetCategory | ComponentCategory | BatteryType | InverterType | EvChargerType
+    TargetCategory
+    | ElectricalComponentCategory
```
llucax, line 208: We also need to keep the old `ComponentCategory` if we want backwards compatibility.

src/frequenz/client/dispatch/types.py
```diff
@@ -181,7 +188,8 @@ def __new__(cls, *categories_input: TargetCategoryInputType) -> Self:
             if isinstance(item, TargetCategory):
                 processed_elements.append(item)
             elif isinstance(
-                item, (ComponentCategory, BatteryType, InverterType, EvChargerType)
+                item,
+                (ElectricalComponentCategory, BatteryType, InverterType, EvChargerType),
```
llucax: Also here, we need to keep the old `ComponentCategory` if we want backwards compatibility.

src/frequenz/client/dispatch/types.py
```diff
@@ -617,15 +625,6 @@ def to_protobuf(self) -> PBDispatch:
         )
 
 
-class Event(Enum):
-    """Enum representing the type of event that occurred during a dispatch operation."""
-
-    UNSPECIFIED = StreamMicrogridDispatchesResponse.Event.EVENT_UNSPECIFIED
-    CREATED = StreamMicrogridDispatchesResponse.Event.EVENT_CREATED
-    UPDATED = StreamMicrogridDispatchesResponse.Event.EVENT_UPDATED
-    DELETED = StreamMicrogridDispatchesResponse.Event.EVENT_DELETED
```
llucax, line 626: We should also re-export the event in common here for backwards compatibility (or leave this class as deprecated).

llucax (commented):
LGTM, but this introduces breaking changes, which we want to avoid as much as possible from now on.

Maybe for this time, since this applies what are hopefully the last (major) breaking changes in the underlying protocol, we could ship the breaking changes, and help users upgrade their code.

llucax (dismissed):

src/frequenz/client/dispatch/_cli_types.py
Marenz: Added

src/frequenz/client/dispatch/types.py
Marenz: added

src/frequenz/client/dispatch/types.py
Marenz: added `ecategory()`

src/frequenz/client/dispatch/types.py
Marenz: added

src/frequenz/client/dispatch/types.py
Marenz: added

src/frequenz/client/dispatch/types.py
Marenz: Just re-exporting is not enough as the member names also changed. I added the new names to this enum for now

Marenz:
* Updated to be backwards compatible and working with 0.3.5 common-api 
* Added a commit to add back the init key parameter for backwards compabilty.

RELEASE_NOTES.md
```diff
@@ -2,14 +2,22 @@
...
+  * `frequenz-api-dispatch` to `v1.0.0-rc3`
+  * `frequenz-client-common` to `v0.3.5`
+  * `grpcio` to `v1.72.1`
 
 ## New Features
 
+* **`ElectricalComponentCategory`**: The client now uses the more specific `frequenz.client.common.microgrid.electrical_components.ElectricalComponentCategory` for targeting components.
+  * The new property `TargetCategory.ecategory` will return an `ElectricalComponentCategory`.
+  * The existing `TargetCategory.category` property is preserved for backward compatibility and returns the corresponding `ComponentCategory`.
+* **New Event Names**: The `dispatch.Event` enum now includes new members prefixed with `EVENT_` (e.g., `EVENT_CREATED`). The old names (`CREATED`, `UPDATED`, `DELETED`) are kept as aliases for backward compatibility.
```
llucax: Why are prefixes added here? We always remove the prefix as they are only needed in protobuf definitions, as all enum values are global, but in python we always remove them. I guess this comes from common now and we need to fix it there, if so I didn't notice it was added with the prefix 😭

RELEASE_NOTES.md
```diff
@@ -2,14 +2,22 @@
...
+* `InverterType.SOLAR` is now deprecated in favor of `InverterType.PV`. The old name can still be used but will be removed in a future release.
+* The following dependencies were updated:
+  * `frequenz-api-dispatch` to `v1.0.0-rc3`
+  * `frequenz-client-common` to `v0.3.5`
+  * `grpcio` to `v1.72.1`
 
 ## New Features
 
+* **`ElectricalComponentCategory`**: The client now uses the more specific `frequenz.client.common.microgrid.electrical_components.ElectricalComponentCategory` for targeting components.
+  * The new property `TargetCategory.ecategory` will return an `ElectricalComponentCategory`.
```
llucax: I would call this `category2`, as recommended in the v0.x.x version guideline. On the first glance I didn't even notice the leading `e`. Also using a _standard_ name make it more obvious that this property should be renamed back to just `category` in the future, in the next breaking release.

src/frequenz/client/dispatch/types.py
llucax: Yeah, I'm not sure why we added the prefixes in common, that was a mistake.

RELEASE_NOTES.md
llucax: I was looking into this and there is no straightforward way to deprecate enum members, but I already have a draft tool to be able to do that. Depending on the urgency for getting this working, for a quick fix we could consider yanking 0.3.5 too and do a 0.3.6 release with the enum without suffixes 😬

Marenz: I would have preferred the non prefix variant too. Let's remove it then in upstream because I think I am the only user anyway

llucax: * https://github.com/frequenz-floss/frequenz-client-common-python/pull/104

RELEASE_NOTES.md
Marenz: Ah, I didn't know it was an actual guideline recommendation too

src/frequenz/client/dispatch/types.py
```diff
@@ -33,10 +33,32 @@
...
+from frequenz.client.common.microgrid.electrical_components import (
+    ElectricalComponentCategory,
+)
+from frequenz.client.common.streaming import Event
 
 from .recurrence import Frequency, RecurrenceRule, Weekday
 
+# Re-export Event for backwards compatibility
+__all__ = [
+    "Event",
```
llucax, line 48: To mark the import location as deprecated you could also instead of re-export it here, add a class like:

```py
@deprecated("Use from frequenz.client.common.streaming.Event instead")
class Event(streaming.Event): pass
```

llucax (dismissed):
Just one last minor optional comment.

src/frequenz/client/dispatch/types.py
Marenz: src/frequenz/client/dispatch/test/_service.py:202: error: Argument "event" to "DispatchEvent" has incompatible type "frequenz.client.common.streaming.Event"; expected "frequenz.client.dispatch.types.Event"  [arg-type]

That doesn't seem to be equivalent at least

Marenz:
Removed tmp commit and will force merge

PR #210, opened 2025-09-02
Fix ci_max after new base-client release

Marenz:
- **Fix call of stream_method in latest base-client**
- **Add Mock for initial_metadata() for base-clients GrpcStreamer**

### DIFF
--- a/src/frequenz/client/dispatch/_client.py
+++ b/src/frequenz/client/dispatch/_client.py
@@ -270,12 +270,9 @@ class DispatchApiClient(BaseApiClient[dispatch_pb2_grpc.MicrogridDispatchService
             request = StreamMicrogridDispatchesRequest(microgrid_id=int(microgrid_id))
             broadcaster = GrpcStreamBroadcaster(
                 stream_name="StreamMicrogridDispatches",
-                stream_method=lambda: cast(
-                    AsyncIterator[StreamMicrogridDispatchesResponse],
-                    self.stub.StreamMicrogridDispatches(
-                        request,
-                        timeout=self._stream_timeout_seconds,
-                    ),
+                stream_method=lambda: self.stub.StreamMicrogridDispatches(
+                    request,
+                    timeout=self._stream_timeout_seconds,
                 ),
                 transform=DispatchEvent.from_protobuf,
                 retry_strategy=LinearBackoff(interval=1, limit=None),
--- a/src/frequenz/client/dispatch/test/_service.py
+++ b/src/frequenz/client/dispatch/test/_service.py
@@ -9,6 +9,7 @@ import logging
 from dataclasses import dataclass, replace
 from datetime import datetime, timezone
 from typing import AsyncIterator
+from unittest.mock import AsyncMock, MagicMock
 
 import grpc
 import grpc.aio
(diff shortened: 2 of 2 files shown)


src/frequenz/client/dispatch/test/_service.py
```diff
@@ -44,6 +44,33 @@
...
+        """Initialize the mock stream.
+
+        Args:
+            stream: The stream to wrap.
+        """
+        self._iterator = stream.__aiter__()
+
+    async def initial_metadata(self) -> None:
+        """Do nothing, just to mock the grpc call."""
+        _logger.debug("Called initial_metadata()")
```
llucax: Why not using a proper `unittest.mock.Mock`? [F]or example](https://github.com/frequenz-floss/frequenz-client-base-python/blob/4afe440e823bd8e888033725ccdc82bd8d1788d9/tests/streaming/test_grpc_stream_broadcaster.py#L59-L68):

```py
def unary_stream_call_mock(
    name: str, side_effect: Callable[[], AsyncIterator[object]]
) -> mock.MagicMock:
    """Create a new mocked unary stream call."""
    # Sadly we can't use spec here because grpc.aio.UnaryStreamCall seems to be
    # dynamic and mock doesn't find `__aiter__` in it when creating the spec.
    call_mock = mock.MagicMock(name=name)
    call_mock.__aiter__.side_effect = side_effect
    call_mock.initial_metadata = mock.AsyncMock()
    return call_mock
```

Then you can do stuff like `m.initial_metadata.return_value = SomeMetadata()`, `m.initial_metadata.assert_called()`, etc.

src/frequenz/client/dispatch/_client.py
```diff
@@ -270,12 +270,9 @@ def _get_stream(
...
-                stream_method=lambda: cast(
-                    AsyncIterator[StreamMicrogridDispatchesResponse],
-                    self.stub.StreamMicrogridDispatches(
-                        request,
-                        timeout=self._stream_timeout_seconds,
-                    ),
+                stream_method=lambda: self.stub.StreamMicrogridDispatches(
+                    request,
+                    timeout=self._stream_timeout_seconds,
                 ),
```
llucax, line 276: Ah, interesting. Maybe other projects could be affected about this, if they didn't remove the casts after moving to use the proper *async stub*. 🤔 

Once more life shows that there are no small or improbable breaking changes 😢

src/frequenz/client/dispatch/test/_service.py
Marenz: Should we use the actual mock library in code outside of test/ ?

llucax: Well, that makes me think why are you mocking outside of tests? :thinking:

llucax: OK, this is the tool we provide for other downstream projects to write tests, right? If this is the case, yes, I think it is pretty appropriate to use `unittest.mock`, maybe unless you expect anyone to use this `disptach.test` package outside tests.

Marenz: Doesn't seem to be very straight forward to mock, your example and some variations of it didn't work

Marenz: Hmm making progress though, let's see

llucax: OK, weird that it works in the client base repo. What kind of issues are you having?

llucax (commented):
LGTM, but now that v0.11.1 is yanked, I would wait until we figure out why is this version failing before we merge this.

PR #213, opened 2025-09-30
Add support for `dispatch_ids` and `queries` filters

Marenz:


src/frequenz/client/dispatch/test/_service.py
```diff
@@ -178,6 +179,44 @@ def _filter_dispatch(
...
+            ):
+                return False
+
+            if not FakeService.matches_query(_filter, dispatch):
+                return False
+
+        return True
+
+    @staticmethod
+    def matches_query(_filter: DispatchFilter, dispatch: Dispatch) -> bool:
```
llucax:
```suggestion
    def matches_query(filter_: DispatchFilter, dispatch: Dispatch) -> bool:
```

Usually a leading `_` is used for *unused* stuff, if you just want to use a keyword or reserved word in any way, a trailing `_` is usually used. If in this case you to shut up pylint, I would add a `pylint: disable=` instead), otherwise the API looks ugly.

src/frequenz/client/dispatch/test/_service.py
```diff
@@ -178,6 +179,44 @@ def _filter_dispatch(
...
+                    try:
+                        query_id = DispatchId(int(query[1:]))
+                        if dispatch.id == query_id:
+                            matches_any_query = True
+                            break
+                    except ValueError:
+                        # not an int, interpret as exact type match (without the #)
+                        if query[1:] == dispatch.type:
+                            matches_any_query = True
+                            break
```
llucax, line 209: To make the error detection more fine-grained:


```suggestion
                    try:
                        int_id = int(query[1:])
                    except ValueError:
                        # not an int, interpret as exact type match (without the #)
                        if query[1:] == dispatch.type:
                            matches_any_query = True
                            break
                    else:
                        query_id = DispatchId(int_id)
                        if dispatch.id == query_id:
                            matches_any_query = True
                            break
```

src/frequenz/client/dispatch/__main__.py
```diff
@@ -291,6 +297,14 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
         # Name of the parameter in client.list()
         filters["target_components"] = target
 
+    # Convert dispatch_ids to iterator if present
```
llucax: It is pretty obvious you are converting to a `iter`, the question (what might be of more value in a comment) is why? :)

tests/test_client.py
```diff
@@ -113,6 +113,121 @@ async def test_list_dispatches(
             assert dispatch == service_side_dispatch
 
 
+# pylint: disable=too-many-locals
```
llucax:
```suggestion
# pylint: disable-next=too-many-locals
```

llucax (dismissed):
I didn't look into tests in much details. All minor comments, if you address them, feel free to force-merge after the push.

src/frequenz/client/dispatch/test/_service.py
Marenz: I probably can make this a private method, too

src/frequenz/client/dispatch/__main__.py
```diff
@@ -292,6 +297,14 @@ async def list_(ctx: click.Context, /, **filters: Any) -> None:
         # Name of the parameter in client.list()
         filters["target_components"] = target
 
+    # Convert dispatch_ids to iterator to match client.list() parameter type
+    if "dispatch_ids" in filters:
+        filters["dispatch_ids"] = iter(filters["dispatch_ids"])
```
llucax, line 302: Again, super minor and irrelevant, but if you can call `iter()` on it, it is already an iterable, right? Why do you need to call it explicitly? It doesn't hurt to do so, so just out of curiosity...

Oh, wait, I just looked at `client.list()` and it takes an iterator instead of an iterable, I see. I think we could change it in the future to `Iterable`, any `Iterator` is iterable too, so it should work. It is pretty annoying not being able to pass a collection (iterable) directly to `list()` and having to explicitly *convert* to an iterator.

Marenz: https://github.com/frequenz-floss/frequenz-client-dispatch-python/pull/223

PR #223, opened 2025-10-13
Fix: Change list method parameters from Iterator to Iterable for better API usability

Marenz:
- Changed target_components, dispatch_ids, and filter_queries parameters
  from Iterator to Iterable in the list method
- This allows users to pass any iterable (list, tuple, generator, etc.)
  instead of requiring specific iterator types
- Improves API flexibility and usability

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,7 @@
 
 ## New Features
 
+* Changed `target_components`, `dispatch_ids`, and `filter_queries` parameters from `Iterator` to `Iterable` in `list` method for better API usability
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -19,3 +19,4 @@
 
 ## Bug Fixes
 
+* Changed `target_components`, `dispatch_ids`, and `filter_queries` parameters from `Iterator` to `Iterable` in `list` method for better API usability
```
llucax: Nitpick, I would put this as a "new feature", nothing was really broken, we are just accepting a wider range of parameters.

llucax (dismissed):

PR #263, opened 2026-02-10
Make `key` the canonical parameter, deprecate `auth_key`

Marenz:
Reverses the previous rename so that `key` / `DISPATCH_API_KEY` are the canonical names again.

- `DispatchApiClient.__init__()`: `key` is the primary parameter, `auth_key` emits a deprecation warning
- CLI: `--api-key` / `DISPATCH_API_KEY` is primary, `--auth-key` / `DISPATCH_API_AUTH_KEY` is deprecated
- Backwards compatible: both names continue to work


llucax (commented):
I would hold on this one too:

1. We need to make sure we agree on whatever name you want to propose, to avoid having to change everything again (including CLI args, constructor parameters and env variables, I would align them ideally, so either `--api-key`/`api_key=`/_API_KEY` or `--key`/`key=`/`_KEY`, but not a mix of both).
2. I would do the change top-down (i.e. change `client-base` first), so it gets more visibility and all projects get updated, otherwise it only adds more confusion if some projects use one and others the other
3. Unrelated, but I would consider removing the env var support, at least for the client library (we can keep it as a CLI-only feature, it makes more sense there). Even when "automatic auth" can be convenient, I guess making auth explicit in code is not a crazy requirement.

Marenz:
> Unrelated, but I would consider removing the env var support, at least for the client library (we can keep it as a CLI-only feature, it makes more sense there). Even when "automatic auth" can be convenient, I guess making auth explicit in code is not a crazy requirement.

It _is_ a CLI only feature, the client/library itself knows nothing about env vars :)

Marenz:
@cwasicki This includes the fix for the epoch timestamp bug you discovered in frequenz-io/frequenz-actor-electricity-trading#570 — passing a `numpy.datetime64` (from `TimeSeries.index`) to `create(start_time=...)` silently produced epoch dispatches. The fix raises `ValueError` immediately for anything that isn't a `datetime` or `"NOW"`, and the corresponding fix in the actor should convert the index timestamps to proper `datetime` objects before constructing `BatteryPowerScheduleEntry`.

cwasicki (commented):
The start time fix LGTM, thank you!

cwasicki (approved):

cwasicki (approved):

